Issue #15, opened 2018-06-20
Implement `POST /matrix` and `GET /matrix?request_id=...`


mweiden:
## Need (user story)
As a user of the Matrix service
I want to query for concatenated matrices given a list of bundle UUIDs.

## Approach

Rough approach (you may want to do this in a streaming fashion).

### `POST /matrix`
1. An endpoint that accepts a list of bundle UUIDs
2. Trigger an async job
3. Return a request id to the user

### async job
1. Query DSS `GET /v1/bundles` for the bundle manifest of each
2. Filter for the matrix files and stream them to a file

### `GET /matrix`
1. Accept a job id from the user
2. Return job status if the job isn't done
3. Return a link to the data if the job is done

## Definition of Done
- [x] A user can create a job to concatenate matrices based on a list of bundle UUIDs
- [x] The API can concatenate matrices from a list of bundle UUIDs and write them to s3
- [x] The user can query for the concatenation job status

PR #16, opened 2018-07-05
[WIP] Enable basic matrix concatenation via the API

l5han:
Fixes https://github.com/HumanCellAtlas/matrix-service/issues/15


chalicelib/request_handler.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @staticmethod
+    def update_request(bundle_uuids, request_id, status):
+        """
+        Create a request status json file in s3 bucket.
+        :param bundle_uuids: A list of bundle uuids.
+        :param request_id: Matrices concatenation request id.
+        :param status: Request status to update.
+        """
+        with open(REQUEST_TEMPLATE_PATH) as f:
```
mweiden: Reading and writing a file to disk every time you update a status is slow. I would hardcode this dict.

test/__init__.py
```diff
@@ -0,0 +1,9 @@
+import logging
+
+formatter = logging.Formatter("%(levelname)s:%(name)s:%(message)s")
+
+stream_handler = logging.StreamHandler()
+stream_handler.setFormatter(formatter)
+
+logger = logging.getLogger("hca-matrix-service-testcase")
+logger.addHandler(stream_handler)
```
mweiden: Here too, I would just use the standard logging configuration as outlined in the AWS doc.

chalicelib/request_handler.py
```diff
@@ -0,0 +1,94 @@
...
+            # Create a request based on a template json file
+            request = json.load(f)
+            request["bundle_uuids"] = bundle_uuids
+            request["status"] = status
+            request["request_id"] = request_id
+
+            if status == RequestStatus.DONE.name:
```
mweiden: It might be simpler simply to pass in and compare the enum itself, rather than the enum's name.

chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        Merge matrices within bundles, and upload the merged matrix to an s3 bucket
+
+        :param bundle_uuids: Bundles' uuid for locating bundles in DSS
+        :param request_id: Merge request id
+        :return:
+        """
+        mtx_paths = self._download_mtx(bundle_uuids)
+        merged_mtx_path = self._concat_mtx(mtx_paths, request_id)
+        self._upload_mtx(merged_mtx_path)
```
mweiden: After you've successfully uploaded the matrix, you should clean up the temp files. I would do this with a `try`+`finally` clause to make sure that the files get cleaned up in case of a failure in subsequent code.

app.py
```diff
@@ -1,171 +1,81 @@
...
+
+mtx_handler = LoomMatrixHandler()
+
+
+@app.route('/')
+def index():
+    return {'hello': 'world'}
+
+
+@app.route('/matrices/concat/{request_id}')
```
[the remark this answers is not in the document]
l5han: Sure. I will do that.

app.py
```diff
@@ -1,171 +1,81 @@
...
+
+app = Chalice(app_name='matrix-service')
+# app.debug = True
+
+mtx_handler = LoomMatrixHandler()
+
+
+@app.route('/')
+def index():
+    return {'hello': 'world'}
```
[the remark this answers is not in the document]
l5han: Will do.

chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,141 @@
...
+        :return: A list of matrices uuids.
+        """
+        mtx_uuids = []
+
+        # Iterate uuids to query DSS for bundle manifest of each
+        for uuid in bundle_uuids:
+            bundle_manifest = self.hca_client.get_bundle(replica="aws", uuid=uuid)
+
+            # Gather up uuids of all the matrix files we are going to merge
+            for file in bundle_manifest["bundle"]["files"]:
```
mckinsel: Consider calling this `file_` or something that's not a keyword.

chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,141 @@
...
+        # Create a temp directory for storing s3 matrix files
+        temp_dir = tempfile.mkdtemp()
+
+        local_mtx_paths = []
+
+        # Download matrices from cloud to a local temp directory
+        for uuid in mtx_uuids:
+            path = os.path.join(temp_dir, uuid + self._extension)
+            with open(path, "wb") as mtx:
+                mtx.write(self.hca_client.get_file(uuid=uuid, replica="aws"))
```
mckinsel: This seems sensible, but the `download` function in the DCP CLI does something quite a bit more complex when it tries to download files from a bundle. I'm guessing it's trying to, among other things, avoid reading the entire file into memory before writing it, but take a look: https://github.com/HumanCellAtlas/dcp-cli/blob/master/hca/dss/__init__.py#L35

chalicelib/matrix_handler.py
ttung: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

chalicelib/request_handler.py
l5han: Done.

app.py
```diff
@@ -1,171 +1,81 @@
...
-    key = query_id + '.loom'
-    with open(out_file, 'rb') as merged_matrix:
-        s3.Bucket(BUCKET_NAME).put_object(Key=key, Body=merged_matrix)
-
-    db = get_db()
-    db.execute('update queries set status="Complete", bucket=?, key=? where query_id = ?',
-               (BUCKET_NAME, key, query_id))
-    db.commit()
 
+            proc = multiprocessing.Process(
```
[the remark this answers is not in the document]
ttung: stepfunctions?

chalicelib/request_handler.py
l5han: If I directly pass in an enum object into the json, then I will have to do work to serialize/deserialize the enum. Is it worth to do that instead of directly passing the enum's name, which is simpler?

chalicelib/matrix_handler.py
l5han: Fixed.

chalicelib/request_handler.py
mweiden: (Discussed this in person)

chalicelib/matrix_handler.py
l5han: Fixed.

chalicelib/__init__.py
```diff
@@ -0,0 +1,14 @@
+import logging
+import uuid
+
+formatter = logging.Formatter("%(levelname)s:%(name)s:%(message)s")
+
+stream_handler = logging.StreamHandler()
+stream_handler.setFormatter(formatter)
+
+logger = logging.getLogger("hca-matrix-service")
+logger.addHandler(stream_handler)
```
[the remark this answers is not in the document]
l5han: fixed.

test/__init__.py
l5han: fixed.

chalicelib/constants.py
```diff
@@ -0,0 +1,19 @@
+import os
+
+# S3 Bucket for storing merged matrices
+MERGED_MTX_BUCKET_NAME = "hca-dcp-matrix-service"
+
+# S3 Bucket for storing matrices concatenation request status
+MERGED_REQUEST_STATUS_BUCKET_NAME = "hca-dcp-matrix-service-request-status"
+
+# S3 Bucket for staging sample matrices
+STAGING_BUCKET = "matrix-service-test"
```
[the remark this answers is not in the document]
l5han: Fixed.

chalicelib/matrix_handler.py
l5han: fixed.

chalice/app.py
```diff
@@ -0,0 +1,80 @@
...
+
+    :param request_id: <string> Matrices concatenation request ID
+    """
+    try:
+        app.log.info("Checking request({}) status.".format(request_id))
+        request_status = RequestHandler.check_request_status(request_id)
+        app.log.info("Request({}) status: {}.".format(request_id, request_status))
+
+        if request_status == RequestStatus.UNINITIALIZED:
+            raise NotFoundError("Request({}) does not exist.".format(request_id))
```
ttung: Because of AWS's consistency model, you will likely run into scenarios where the request status will report uninitialized even though it's not really the case.  You could solve the problem by never GET'ing a file before you PUT it, but it adds a bit of complexity to your design.

https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html#ConsistencyModel

ttung:
sorry, i’m still subscribed to your diff so i can’t help but comment:

re https://github.com/HumanCellAtlas/matrix-service/commit/c6d3de1: consider using https://github.com/chanzuckerberg/cloud-blobstore?  it’s what we use in the data-store, and encompasses all your functionality.  it's published in pypi.

PR #17, opened 2018-07-13
Add vendor directory

mweiden:


ttung:
IMO you should only vendor things that have machine-architecture-specific code.

mweiden:
@ttung that makes sense. I didn't scrutinize the contents. We'll fix this.

PR #21, opened 2018-07-20
Enable basic matrix concatenation via the API

l5han:


Makefile
```diff
@@ -0,0 +1,68 @@
...
+	. venv/bin/activate && pip install -r requirements-dev.txt --upgrade
+
+.PHONY: test
+test:
+	. venv/bin/activate && PYTHONPATH=chalice \
+		python -m unittest discover -s tests -p '*_tests.py'
+
+.PHONY: build
+build:
+	@read -p "Enter AWS_PROFILE: " aws_profile; \
```
ttung: consider using AWS secrets manager, and putting all of this config in a script that someone can run once.

mweiden: @l5han you can use this code https://github.com/HumanCellAtlas/dcplib/blob/master/dcplib/aws_secret.py

chalice/app.py
```diff
@@ -0,0 +1,128 @@
...
+    return {"request_id": request_id}
+
+
+def ms_sqs_queue_listener(event, context):
+    """
+    Create a lambda function that listens for the matrix service's SQS
+    queue events.Once it detects an incoming message on the queue, it
+    will process it by launching a job to do the matrices concatenation
+    based on the message content.
+    :param event: SQS Queue event.
```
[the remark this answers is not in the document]
l5han: Yes. It is already working on AWS. Its name is "matrix-service-sqs-listener".

chalice/chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,172 @@
...
+
+                # Update the request status to DONE
+                RequestHandler.update_request(
+                    bundle_uuids=bundle_uuids,
+                    request_id=request_id,
+                    job_id=job_id,
+                    status=RequestStatus.DONE,
+                    time_spent_to_complete="{} seconds".format(end_time - start_time)
+                )
+            except (SwaggerAPIException, OSError, Exception):
```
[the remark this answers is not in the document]
l5han: When timeout, the request message will be put back to the SQS queue, and waiting for a retry. After 3rd retry fails, it will be direct from the SQS Queue to a Dead Letter Queue. Right now, there is no further handling for the messages on the dead letter queue, but I intend to have some mechanism so that those requests will be set to "ABORT" status.

chalice/chalicelib/config.py
```diff
@@ -0,0 +1,71 @@
...
+CONFIGURATION_FILE = os.path.join(BASE_DIR, "config.json")
+
+# Load configuration
+with open(CONFIGURATION_FILE) as f:
+    configuration = json.load(f)
+    hca_client_host = configuration["hca_host"]
+    ms_secret_name = configuration["ms_secret_name"]
+
+# Default directory for all temp files
+TEMP_DIR = "/tmp"
```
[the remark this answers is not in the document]
l5han: Yes, I do use `tempfile`. The thing is in lambda, I am only allowed to write to a directory called `tmp/`. Thus, even if I use `tempfile` to create files or directories, I have to specify the root directory for them is `tmp/`, which is where this variable `TEMP_DIR` being used for.

chalice/app.py
```diff
@@ -0,0 +1,128 @@
...
+        # Send the request to sqs queue if the request has been abort before
+        if merge_request_status == RequestStatus.ABORT:
+            SqsQueueHandler.send_msg_to_ms_queue(bundle_uuids=bundle_uuids, request_id=request_id)
+
+    except BlobNotFoundError:
+        try:
+            # Send the request to sqs queue if the request has not been made before
+            SqsQueueHandler.send_msg_to_ms_queue(bundle_uuids=bundle_uuids, request_id=request_id)
+        except BlobStoreUnknownError:
+            return BadRequestError(traceback.format_exc())
```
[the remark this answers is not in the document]
l5han: You mean I need to raise it so that it can be shown in the cloudwatch log?

chalice/app.py
```diff
@@ -0,0 +1,128 @@
...
+    bundle_uuids = request.json_body
+    request_id = RequestHandler.generate_request_id(bundle_uuids)
+
+    logger.info("Request ID({}): Received request for concatenating matrices within bundles {};"
+                .format(request_id, str(bundle_uuids)))
+
+    try:
+        merge_request = RequestHandler.get_request(request_id=request_id)
+        merge_request_body = json.loads(merge_request)
+        merge_request_status = merge_request_body["status"]
```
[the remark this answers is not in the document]
l5han: Sounds a good idea. I will change it accordingly.

chalice/app.py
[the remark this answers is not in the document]
l5han: Yep. It seems like `AWS Lambda Supports Amazon SQS as an Event Source` starts in June 28, 2018.

chalice/app.py
```diff
@@ -0,0 +1,128 @@
+import json
+import traceback
+
+from chalice import Chalice, NotFoundError, BadRequestError
+from hca.util import SwaggerAPIException
+from cloud_blobstore import BlobNotFoundError, BlobStoreUnknownError
```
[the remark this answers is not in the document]
mckinsel: Sure, but here's the implementation of `BlobNotFoundError`:

```python
class BlobStoreError(Exception):
    pass

class BlobNotFoundError(BlobStoreError):
    pass
```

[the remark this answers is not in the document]
l5han: Okay. I did not notice this before. But even though it is just a subclass of `Exception`, I still have to import it right? Otherwise, I cannot know that the exception is caused by not finding a key in the s3 bucket by just catching a broad `Exception`. Let me know if you have some better ideas for handling it.

chalice/app.py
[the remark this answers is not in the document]
mckinsel: Well, you can return an `Exception`, but it's hardly ever what you want to do. I assume here you want to trigger exception handling behavior, no try to get chalice to serialize and return `BadRequestError` in a response to the user.

chalice/chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,165 @@
...
+        ))
+
+        with open(path, "rb") as merged_matrix:
+            try:
+                s3_blob_store.upload_file_handle(
+                    bucket=MERGED_MTX_BUCKET_NAME,
+                    key=request_id + self._suffix,
+                    src_file_handle=merged_matrix
+                )
+            except Exception as e:
```
mckinsel: The
```python
try:
    ...
except Exception as e:
    raise e
```
pattern appears a number of times, but it doesn't really do anything does it? I would omit `try`/`except` block for these. The behavior is the same, and it won't look like there's special exception handling happening.

chalice/app.py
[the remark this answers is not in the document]
mckinsel: Well, you could just stick your own `Exception` subclasses in `__init__` (or wherever makes sense). You don't need to pull in a dependency for for that.

That said, I hadn't noticed that you're using `cloud_blobstore.s3.s3_blob_store`, which is more than just a bare `Exception` subclass, so it's probably fine.

chalice/chalicelib/request_handler.py
```diff
@@ -0,0 +1,128 @@
...
+        request["bundle_uuids"] = bundle_uuids
+        request["status"] = status.name
+        request["request_id"] = request_id
+        request["job_id"] = job_id
+        request["time_spent_to_complete"] = time_spent_to_complete
+        request["reason_to_abort"] = reason_to_abort
+
+        if status == RequestStatus.DONE:
+            # Key for merged matrix stored in s3 bucket
+            mtx_key = request_id + ".loom"
```
mckinsel: I think a reference to the loom extension is a little out of place in `RequestHandler`.

chalice/chalicelib/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+def get_mtx_paths(dir: str, mtx_suffix: str) -> List[str]:
+    """
+    Get all matrices file paths within a directory.
+    :param dir: Directory that contains matrix files.
+    :param mtx_suffix: Suffix for the matrix files.
+    :return: A list of mtx that ends with the specific suffix in the directory.
+    """
+    mtx_paths = []
+
+    for dname, _, fnames in os.walk(dir):
```
mckinsel: I think you can use `pathlib` for this now, something like `pathlib.Path(dir).glob("**/*." + mtx_suffix)`. Up to you, but the `os.walk` syntax always looked super ugly to me.

chalice/chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,165 @@
...
+        try:
+            logger.info("Concatenate matrices for request_id: {}".format(request_id))
+
+            # Print out the current usage of /tmp/ directory
+            call(["df", "-H", TEMP_DIR])
+
+            logger.info("tmp directory contains: {}".format(str(os.listdir(TEMP_DIR))))
+
+            # Clean /tmp folder before each run
+            call('rm -rf /tmp/*', shell=True)
```
mckinsel: You should use `TEMP_DIR` here.

chalice/chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,165 @@
...
+    Matrix handler for .loom file format
+    """
+
+    def __init__(self) -> None:
+        super().__init__(".loom")
+
+    def _concat_mtx(self, mtx_paths: List[str], out_file: str) -> None:
+        try:
+            logger.info("Combining matrices to %s.", out_file)
+            loompy.combine(mtx_paths, out_file)
```
mckinsel: Looks like there's a `batch_size` argument to this that we could try cranking up.

chalice/chalicelib/config.py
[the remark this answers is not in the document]
mckinsel: Okay cool, sounds fine to me. I only ask because I'm not quite sure what the exact promise made by lambda is. Is `/tmp` writable or is it, say, the value of `$TMPDIR`.

chalice/chalicelib/matrix_handler.py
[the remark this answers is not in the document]
mckinsel: Sounds like a plan, thanks.

chalice/chalicelib/request_handler.py
```diff
@@ -0,0 +1,128 @@
...
+            RequestHandler.get_request_body(request_id=request_id)
+        except BlobNotFoundError:
+            with tempfile.TemporaryFile(dir=TEMP_DIR, suffix=JSON_SUFFIX) as f:
+                s3_blob_store.upload_file_handle(
+                    bucket=REQUEST_STATUS_BUCKET_NAME,
+                    key=request_key,
+                    src_file_handle=f
+                )
+        except Exception as e:
+            raise e
```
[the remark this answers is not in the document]
mckinsel: Actually I think the effects of this are kind of severe. Making a GET for the object here means you don't get read-after-write consistency even on the first PUT.

chalice/chalicelib/request_handler.py
```diff
@@ -0,0 +1,128 @@
...
+
+            mtx_url = "s3://{}/{}".format(
+                MERGED_MTX_BUCKET_NAME,
+                mtx_key
+            )
+            request["merged_mtx_url"] = mtx_url
+
+        try:
+            s3_request_obj = RequestHandler.s3.Object(REQUEST_STATUS_BUCKET_NAME, request_key)
+            s3_request_obj.put(Body=json.dumps(request))
```
mckinsel: I'm pretty sure S3 only offers eventual consistency for PUTs that overwrite an existing file. This will probably lead to lots of weird bugs.

l5han: Hey Marcus. Can you please elaborate more this issue? I know that S3 offers eventual consistency model, but I do not see any potential bug that could arise here. Can you give me an example? Thanks.

chalice/chalicelib/matrix_handler.py
l5han: Okay. Do you have any recommended value that I can use for `batch_size`? Or do you think I can conduct some experiments on it to find the most optimal one?

chalice/chalicelib/matrix_handler.py
l5han: Thanks for your reminding. I will fix it.

chalice/chalicelib/__init__.py
l5han: Sure. I will switch to use pathlib.

chalice/chalicelib/request_handler.py
l5han: Good catch. I will find a more elegant way to do this.

chalice/chalicelib/config.py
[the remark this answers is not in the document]
l5han: It is `/tmp`, not the value of `$TMPDIR`. I have done some experiments on it, and they work correctly.

chalice/app.py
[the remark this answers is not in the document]
l5han: Okay. Cool.

chalice/chalicelib/matrix_handler.py
l5han: Okay. I will modify it to make the code cleaner.

chalice/chalicelib/matrix_handler.py
mckinsel: Probably need to experiment a little. I just recall that you were a little surprised at the speed of the merges, so it might be worth trying to increase it. I can say anecdotally that increasing the size/reducing the number of reads and writes can improve HDF5 I/O a lot.

chalice/chalicelib/request_handler.py
mckinsel: Well, I think you've already thought about one case a bit in this comment:

          2. Before receiving the msg from SQS queue, a duplicate mtx
             concatenation request has been initialized. As a result,
             the job id in the status file in s3 has been updated to
             a new value.

The problem is you don't know which version of the status file in S3 you're going to get. It could be the new one written by the duplicate matrix request, or it could be the first one with a job_id that matches what's in the queue.

l5han: Okay. If it gets the new one written by the duplicate matrix request, it will find that there is a inconsistency between the job id it receives and the one in the S3 file. As a result, it will abandon processing a merge on it. This behavior is expected.

If it gets the old one instead of the new one. It will merge the matrices because of the match in job ids. And, when the duplicate request being polled from the queue and processed by the lambda function, since there is still a match in job ids, that requests will still be handled. As a result, we do a matrices concatenation twice.

Thus, it seems like that in the worst case, the AWS S3 consistency model will cause the current version of matrix service not being able to avoid duplicate matrices concatenation. Is my understanding correct?

chalice/chalicelib/matrix_handler.py
l5han: Sure. I will do that.

chalice/chalicelib/request_handler.py
mckinsel: Hmm, yeah this is good to think through. Consider the follow though:

- Two nearly simultaneous POSTs to /concat get made with the same list of bundle ids
- Neither sees an object named request_id_1.json (making up a name here). So each generates a different job_id, each PUTs an object with a different job_id into S3 and a message into the queue. Call these request_id_1.json[job_id_1], request_id_1.json[job_id_2], message[job_id_1], and message[job_id_2].

Now the queue triggers two lambda executions, one for each message. 

- Suppose the first lambda gets message[job_id_1]. It then looks at S3 for request_id_1.json. We don't know if it's going to get request_id_1.json[job_id_1] or request_id_1.json[job_id_2]. But suppose it gets the latter. It's going to see that the message and S3 job_ids don't match, and it's going to halt.
- Now the second lambda gets message[job_id_2]. It also looks at S3 for request_id_1.json, and we still don't know which version it's going to get, but suppose we're unlucky and it gets request_id_1.json[job_id_1]. Just like the first lambda it also halts.

Now we have a matrix request in a permanent INITIALIZED state, but nothing is happening. I think this set of events is possible, but I might be overlooking something.

More generally, it's worthing looking through the code at every `get_request_field` or `get_request_body` and considering what could happen if you got a stale or inconsistent S3 object.

Also, I don't think this needs to block merging this PR. But if we decide we need to address it, we should merge and file an issue.

l5han: I see your point. I will take carefully look at every operation related to `get` things from S3.

chalice/chalicelib/request_handler.py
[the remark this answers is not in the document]
l5han: Yes, that's true. But even if I don't get read-after-write consistency on the first PUT, the worst thing that could happen is: 

 * Suppose two consecutive `update_request()` happens, and both of them having the same request id.  
 * The first `update_request()` GET the object from S3 and receives 404. Thus, it PUT a new status file into S3.
 * When the second `update_request()` tries GET the object from S3, (suppose the first PUT has not fully propagated) due to the eventual consistency, the second request stills receives 404. As a result, it will also PUT a new file into S3.

The main point of having a GET before PUT here is to avoid create redundant files in the local disk as there is a limitation on the disk size I can use for each lambda execution. Under the scenario I describe above, even though there are still redundant local files being created, it will not cause any bug, right? 

Or, do I miss something?

chalice/app.py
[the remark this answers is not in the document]
l5han: Hmm. I do not think about trigger any exception handling behavior here. If there is an exception happens when the service sends a message to SQS queue, there is no way for me handle that, right? For example, the exception might be caused by the SQS queue does not exist (Suppose the SQS queue is accidentally deleted by someone). When those things happen, I can only tell users there is something wrong in the internal server. It think the service should return 500 internal server error instead of 400 bad request error in here.

chalice/chalicelib/request_handler.py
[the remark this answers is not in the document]
mckinsel: My main concern here is that we're not getting anything out of this code block, and the function would work the same if we just deleted it. I think all this does is upload an empty file to S3, right? What's the benefit of that? If there aren't any benefits, then we don't need to investigate the costs too much.

chalice/app.py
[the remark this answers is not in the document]
mckinsel: Errr, look three lines down. There you _raise_ `BadRequestError`, but here you _return_ it. Those are different things. Why are you returning this one but raising the other?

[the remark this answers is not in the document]
l5han: It is a typo. Both of them should be raised in order for chalice to correctly return the error messages. Sorry about that.

chalice/chalicelib/request_handler.py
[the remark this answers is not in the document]
l5han: If I delete it, and directly create and upload a request status file to S3 each time when `update_request()` is called, then the following problem can happen:  

Suppose there is a request which asks for merging N matrices,  and when we download kth matrix, where k <= N, there is no more disk space available for that lambda invocation. As a result, the `OSError` will be caught in the following snippet of code:

```python
except Exception as e:
    # Update the request status to ABORT
     RequestHandler.update_request(
                matrix_format=self.suffix,
                bundle_uuids=bundle_uuids,
                request_id=request_id,
                job_id=job_id,
                status=RequestStatus.ABORT,
                reason_to_abort=traceback.format_exc()
      )

      raise e

However, since `update_request()` will attempt to create a new file, but there is actually no more space on the disk, the call to `update_request()` will cause another exception without setting the the request status to `ABORT`.

File: chalice/app.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (mckinsel) [reply chalice/app.py]: Okay great, that's what I figured.

File: chalice/chalicelib/request_handler.py (hunk 13 above)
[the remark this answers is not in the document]
Comment (mckinsel) [reply chalice/chalicelib/request_handler.py]: I don't think I understand. I'm pretty sure `update_request` doesn't do any disk I/O except for the portion I'm proposing to delete.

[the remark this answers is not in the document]
Comment (l5han) [reply chalice/chalicelib/request_handler.py]: I am so sorry. It seems like I have a huge misunderstanding here. I originally think that I can only upload a file to S3 as a key. But I find that 

```python
boto3.resource("s3").Object()

will create one for me if it does not exist. I think you are totally right. I will delete that block of code.

File: chalice/app.py [hunk 15]
```diff
@@ -0,0 +1,147 @@
...
+    :param request_id: Matrices concatenation request ID.
+    """
+    try:
+        merge_request_body = RequestHandler.get_request_body(request_id=request_id)
+        return merge_request_body
+
+    except BlobNotFoundError:
+        raise NotFoundError("Request({}) has not been initialized.".format(request_id))
+
+    except:
```
ttung: out of the scope of this diff, but it might make sense to mark all your routes with a decorator that serves as a common fallback exception handler.  this way, you can produce error messages with a consistent format.

chalice/app.py
```diff
@@ -0,0 +1,147 @@
...
+@app.route('/matrices/concat', methods=['POST'])
+def concat_matrices():
+    """
+    Concat matrices within bundles based on bundles uuid
+    """
+    request = app.current_request
+    bundle_uuids = request.json_body
+    request_id = RequestHandler.generate_request_id(bundle_uuids)
+
+    logger.info("Request ID({}): Received request for concatenating matrices within bundles {};"
```
ttung: python 3.6+?  you can use f-style strings.

chalice/app.py
```diff
@@ -0,0 +1,147 @@
...
+    a request ID.
+
+    :param request_id: Matrices concatenation request ID.
+    """
+    try:
+        merge_request_body = RequestHandler.get_request_body(request_id=request_id)
+        return merge_request_body
+
+    except BlobNotFoundError:
+        raise NotFoundError("Request({}) has not been initialized.".format(request_id))
```
ttung: TODO: is subject to race conditions on S3 PUT

chalice/chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,165 @@
...
+                            .format(bundle_uuid, dest_name))
+
+                hca_client.download(
+                    bundle_uuid=bundle_uuid,
+                    replica="aws",
+                    dest_name=dest_name,
+                    metadata_files=(),
+                    data_files=("*{}".format(self._suffix),)
+                )
+            except Exception as e:
```
ttung: ??

you can just remove the try-block then.

chalice/chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,165 @@
...
+
+        :param bundle_uuids: Bundles' uuid for locating bundles in DSS.
+        :param request_id: Merge request id.
+        :param job_id: Job id of the request.
+        """
+        try:
+            logger.info("Concatenate matrices for request_id: {}".format(request_id))
+
+            # Print out the current usage of /tmp/ directory
+            call(["df", "-H", TEMP_DIR])
```
ttung: shelling out is always dicey,  i would avoid it if possible.

chalice/chalicelib/matrix_handler.py
ttung: https://docs.python.org/3/library/shutil.html#shutil.rmtree

chalice/chalicelib/request_handler.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @staticmethod
+    def generate_request_id(bundle_uuids: List[str]) -> str:
+        """
+        Generate a request id based on a list of bundle uuids.
+        :param bundle_uuids: A list of bundle uuids.
+        :return: Request id.
+        """
+        bundle_uuids.sort()
+        m = hashlib.sha256()
```
ttung: If you want to use a checksum to uniquely identify a run to avoid doing it again, you will need bundle fqids.  uuids is insufficient, as you can get a different version on a subsequent run.

chalice/chalicelib/matrix_handler.py
l5han: Marcus also points that out. I will remove it.

chalice/chalicelib/matrix_handler.py
l5han: Okay. I will replace it with `os.statvfs()`.

chalice/chalicelib/matrix_handler.py
l5han: I will replace it.

chalice/app.py
l5han: I will switch to use f-style strings.

chalice/chalicelib/request_handler.py
l5han: Hey Marcus. Do you think enable S3 versioning can help to address this problem. Whenever we try to update the request status file, we always create a new version of it instead of overwriting the old one. And, when we try to GET object from S3, we always get the newest version of it. Doing so can guarantee the `read-after-write consistency` provided by S3 because there is no overwrite happens.

chalice/app.py
l5han: Thanks for this good advice. I will do that.

chalice/chalicelib/request_handler.py
mckinsel: I don't think so. Specifically, I don't think this is true: "And, when we try to GET object from S3, we always get the newest version of it." The S3 consistency model stays the same; you just can refer to an object that otherwise would have been overwritten. The solution is probably to switch to an AWS product with better consistency guarantees. DynamoDB might be enough.

chalice/chalicelib/matrix_handler.py
```diff
@@ -0,0 +1,187 @@
...
+                keywords = {
+                    "bundle_uuid": job,
+                    "replica": "aws",
+                    "dest_name": dest_name,
+                    "metadata_files": (),
+                    "data_files": (f'*{self.suffix}',)
+                }
+
+                hca_download_workers[job] = HcaDownloadWorker(**keywords)
+                hca_download_workers[job].start()
```
mckinsel: This all seems like a lot of code to parallelize a function call over a list. Could you use like `concurrent.futures.ThreadPoolExecutor.map` or `multiprocessing.pool.ThreadPool.map`?

l5han: That's what I tried to do at the beginning. However, due to the Lambda execution environment not having `/dev/shm` (shared memory for processes) support, I can’t use multiprocessing.Queue or multiprocessing.Pool.

mckinsel: Right, but what about the two methods I suggested? You don't have multiple processors in a lambda, and I'm pretty sure you're I/O bound here anyway. So why not use threads?

l5han: Sounds reasonable. I will try using threads.

chalice/app.py
```diff
@@ -15,31 +15,39 @@
 mtx_handler = LoomMatrixHandler()
 
 
+def api_endpoint_decorator(api_endpoint_func):
+    @wraps(api_endpoint_func)
+    def api_endpoint_wrapper(*args, **kwargs):
+        try:
+            return api_endpoint_func(*args, **kwargs)
+        except KeyError:
```
mckinsel: What are you trying to catch here?

chalice/app.py
```diff
@@ -0,0 +1,153 @@
...
+from chalicelib.sqs import SqsQueueHandler
+
+app = Chalice(app_name='matrix-service-api')
+app.debug = True
+
+# Replace handler here for supporting concatenation on other matrix formats
+mtx_handler = LoomMatrixHandler()
+
+
+def api_endpoint_decorator(api_endpoint_func):
```
mckinsel: I know this was a request from @ttung, so I'm curious what he thinks. But, this just collapses every possible error to a 500 with a traceback. That doesn't seem like an improvement 😕

chalice/chalicelib/sqs.py
```diff
@@ -0,0 +1,65 @@
...
+        job_id = rand_uuid()
+
+        logger.info(f'Request ID({request_id}): Initialize the request with job id({job_id})')
+
+        RequestHandler.put_request(
+            bundle_uuids=bundle_uuids,
+            request_id=request_id,
+            job_id=job_id,
+            status=RequestStatus.INITIALIZED
+        )
```
mckinsel: What if you made this a [conditional write](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html#WorkingWithItems.ConditionalUpdate)? That might resolve the issue of identical requests getting put into the queue.

l5han: Thanks for pointing that out. I will use a conditional write here.

chalice/app.py
l5han: If the `api_endpoint_func()` is `check_request_status()`, then an `KeyError` exception indicates that the request has not been made before.

mckinsel: Ah gotcha, that makes sense. In the future we might want to make a special exception for that, but don't need to worry about that for now.

chalice/chalicelib/request_handler.py
mweiden: @ttung This is an important point. I am going to add a ticket for this, but may be added in a subsequent PR

chalice/app.py
mweiden: @mckinsel I believe the way this is handled in the DSS is to have custom exception types that can be transformed into the right responses. This would not be hard to fix if we lift and modify this code.

https://github.com/HumanCellAtlas/data-store/blob/master/dss/error.py

Let's make a separate issue for this and submit it in a different PR.

mckinsel:
Hey let's merge it! Thanks @l5han, this was a ton of work!

PR #28, opened 2018-08-27
Add initial mock API for orange box testing

mckinsel:
This a PR for some concrete discussion of the API the data portal will use to interact with the matrix service.

### DIFF
--- a/dummy-api/matrix-service-api.yaml
+++ b/dummy-api/matrix-service-api.yaml
@@ -1,0 +1,106 @@
+swagger: '2.0'
+basePath: '/v0'
+info:
+  title: HCA Matrix Service
+  version: 0.0.0-dummy
+schemes:
+  - https
+consumes:
+  - application/json
+produces:
+  - application/json
+definitions:
+  MatrixRequest:
+    type: object
+    properties:
+      bundle_fqids:
+        type: array
+        items:
+          type: string
+          format: dcp-fqid
+      bundle_fqids_url:
+        type: string
+        format: url
+      format:
+        $ref: '#/definitions/MatrixFormat'
+  MatrixFormat:
+    type: string
+    enum:
+      - zarr
+      - loom
+      - csv
+  MatrixResponse:
+    type: object
+    properties:
+      request_id:
+        type: string
+        format: uuid
+      status:
+        $ref: '#/definitions/MatrixRequestStatus'
+      key:
+        type: string
+      eta:
+        type: string
+      message:
+        type: string
+      links:
+        type: array
+        items:
+          type: object
+  MatrixRequestStatus:
+    type: string
+    enum:
+      - In Progress
+      - Complete
+      - Failed
+paths:
+  /matrix:
+    post:
+      x-amazon-apigateway-integration:
+        type: aws_proxy
+        httpMethod: POST
+        uri:
+          Fn::Join:
+            - ""
+            - - "arn:aws:apigateway:"
+              - Ref: "AWS::Region"
+              - ":lambda:path/2015-03-31/functions/"
+              - Fn::GetAtt: [MatrixPostFunction, Arn]
+              - "/invocations"
+      parameters:
+        - in: body
+          name: body
+          schema:
+            $ref: '#/definitions/MatrixRequest'
+      responses:
+        "202":
+          description: "matrix request accepted"
+          schema:
+            $ref: '#/definitions/MatrixResponse'
+        "400":
+          description: "bad request"
+  /matrix/{request_id}:
+    get:
+      x-amazon-apigateway-integration:
+        type: aws_proxy
+        httpMethod: POST
+        uri:
+          Fn::Join:
+            - ""
+            - - "arn:aws:apigateway:"
+              - Ref: "AWS::Region"
+              - ":lambda:path/2015-03-31/functions/"
+              - Fn::GetAtt: [MatrixGetFunction, Arn]
+              - "/invocations"
+      parameters:
+        - in: path
+          name: request_id
+          type: string
+          required: true
+      responses:
+        "200":
+          description: "request found"
+          schema:
+            $ref: '#/definitions/MatrixResponse'
+        "404":
+          description: "request not found"
--- a/dummy-api/deploy.sh
+++ b/dummy-api/deploy.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+set -exo pipefail
+
+echo "Creating stack named $1"
+echo "Uploading to bucket $2"
+
+AWS="aws"
+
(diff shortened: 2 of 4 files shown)


dummy-api/matrix-service-api.yaml
```diff
@@ -0,0 +1,115 @@
...
+              - ":lambda:path/2015-03-31/functions/"
+              - Fn::GetAtt: [MatrixPostFunction, Arn]
+              - "/invocations"
+      parameters:
+        - in: body
+          name: body
+          schema:
+            $ref: '#/definitions/MatrixRequest'
+      responses:
+        "201":
```
brianraymor: Since this is an asynchronous operation, I would suggest a [202 (Accepted)](https://tools.ietf.org/html/rfc7231#section-6.3.3):

* [Azure Best Practices](https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design#asynchronous-operations)
* [RedHat learnings](https://restful-api-design.readthedocs.io/en/latest/methods.html#asynchronous-requests)

dummy-api/matrix-service-api.yaml
```diff
@@ -0,0 +1,115 @@
...
+    type: object
+    properties:
+      request_id:
+        type: string
+        format: uuid
+      status:
+        $ref: '#/definitions/MatrixRequestStatus'
+      key:
+        type: string
+  MatrixRequestStatus:
```
brianraymor: [Azure Best Practices](https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design#asynchronous-operations)

_If the client sends a GET request to this endpoint, the response should contain the current status of the request. **Optionally, it could also include an estimated time to completion or a link to cancel the operation**_

`{`
   ` "status":"In progress",`
    `"link": { "rel":"cancel", "method":"delete", "href":"/api/status/12345" }`
`}`

dummy-api/matrix-service-api.yaml
```diff
@@ -0,0 +1,115 @@
...
+    type: string
+    enum:
+      - Initialized
```
brianraymor: What is the difference between the _Initialized_ and _Running_ state from an application perspective? Perhaps replace with _Pending_ or _In Progress_?

dummy-api/matrix-service-api.yaml
```diff
@@ -0,0 +1,115 @@
...
+      - Running
+      - Complete
+      - Failed
```
brianraymor: How does an application diagnose _Failed_ state?

dummy-api/matrix-service-api.yaml
```diff
@@ -0,0 +1,115 @@
+swagger: '2.0'
+info:
+  title: HCA Matrix Service
+  version: 0.0.0-dummy
+schemes:
+  - http
```
brianraymor: Based on https://github.com/HumanCellAtlas/dcp/issues/102 - I don't think that the http scheme should be supported.

dummy-api/matrix-service-api.yaml
```diff
@@ -0,0 +1,115 @@
+swagger: '2.0'
```
brianraymor: Do we want to specify _BasePath_ - perhaps /v0 ?

dummy-api/matrix-service-api.yaml
mckinsel: Yep, will change.

dummy-api/matrix-service-api.yaml
mckinsel: So I can add a placeholder field for estimated time or some other value that represents the scale of the request. I'm a bit concerned that the orange box will depend on it prematurely though. It may not convey useful information in our v0 release.

dummy-api/matrix-service-api.yaml
mckinsel: No difference I suppose. I'll convert to `In Progress`.

dummy-api/matrix-service-api.yaml
mckinsel: Yeah. The azure doc you've linked suggests there are a few schemes to accomplish this, but basePath is the one I'm most familiar with.

dummy-api/matrix-service-api.yaml
mckinsel: 👍 yeah want to follow whatever the dcp's doing for this.

dummy-api/matrix-service-api.yaml
mckinsel: So I added a "message" to the response. Think that's enough?

dummy-api/matrix-service-api.yaml
```diff
@@ -0,0 +1,124 @@
...
+        type: string
+        format: url
+      filter:
+        $ref: '#/definitions/MatrixFilter'
+      format:
+        $ref: '#/definitions/MatrixFormat'
+  MatrixFilter:
+    type: object
+    properties:
+      pandas_filter_string:
```
brianraymor: What is the actual format/content of the _pandas_filter_string_?

NoopDog:
Yes, this looks like what we spoke about in our meetings and is something we can work with. I need also to checkout the pandas filter and see what kind of options we might offer. 

I see we can send the bundle UUIDs directly instead of sending a URL to S3. That may be helpful go get us end to end more easily.

NoopDog:
If we could get some kind of progress indicator in the response that would be helpful does not need to map directly to time, but maybe just the "bundles processed" 2/15,000, 3/15,000 etc might be nice so its apparent there is some progress being made.

NoopDog:
Is this ready to send over some posts? What environment (dev/integration/staging) should the bundle Ids be from?

NoopDog:
Regarding the API/Key is this required or can this be disabled? Currently the data is all open and anyone can hit the data browser and select and download  files without authenticating. Do we need the key on this API and if so, is it unique per user if so how do we get the user their key?

dummy-api/matrix-service-api.yaml
```diff
@@ -0,0 +1,124 @@
...
+        type: string
+  MatrixFormat:
+    type: string
+    enum:
```
brianraymor: This should be reduced to the set of formats supported in v0. Based on today's discussion that would be:
-zarr
-loom
-csv _which is an addition_

dummy-api/matrix-service-api.yaml
```diff
@@ -0,0 +1,124 @@
...
+      - In Progress
+      - Complete
+      - Failed
+paths:
+  /matrix:
+    post:
+      security:
+        - api_key: []
```
brianraymor: Based on the Sept 17 discussion, the **api_key** can also be removed for v0.

PR #45, opened 2018-10-02
fix linter issues

calvinnhieu:
- exclude modules from lint rule
- add flake8 config
- fix linter issues


calvinnhieu:
Thanks @parthshahva!

calvinnhieu:
@mweiden thanks :+1:

PR #48, opened 2018-10-04
update test dir structure to match matrix/

calvinnhieu:


calvinnhieu:
@parthshahva Not sure if you started writing unit tests that this would affect

PR #52, opened 2018-10-05
Create lambda handler class

calvinnhieu:
New class: `LambdaHandler` is responsible for handling all interactions with AWS lambda. Currently the handler only provides an `invoke` function.

New enum: `LambdaName` captures the resource names of matrix service Lambdas

### DIFF
--- a/tests/unit/common/test_lambda_handler.py
+++ b/tests/unit/common/test_lambda_handler.py
@@ -1,0 +1,23 @@
+import unittest
+
+from botocore.stub import Stubber
+
+from matrix.common.lambda_handler import LambdaHandler
+from matrix.common.lambda_handler import LambdaName
+
+
+class TestLambdaHandler(unittest.TestCase):
+    """
+    Environment variables are set in tests/__init__.py
+    """
+    def setUp(self):
+        self.handler = LambdaHandler()
+        self.mock_handler = Stubber(self.handler._client)
+
+    def test_invoke(self):
+        expected_params = {'FunctionName': LambdaName.DRIVER.value,
+                           'InvocationType': "Event",
+                           'Payload': b"{}"}
+        self.mock_handler.add_response('invoke', {}, expected_params)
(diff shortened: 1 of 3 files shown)


calvinnhieu:
TODO 10/5: write tests

calvinnhieu:
Sg. Will start with this one

calvinnhieu:
Added tests

tests/unit/common/test_lambda_handler.py
```diff
@@ -0,0 +1,20 @@
...
+        self.handler = LambdaHandler()
+        self.mock_handler = Stubber(self.handler._client)
+
+    def test_invoke(self):
+        expected_params = {'FunctionName': LambdaName.DRIVER.value,
+                           'InvocationType': "Event",
+                           'Payload': b"{}"}
+        self.mock_handler.add_response('invoke', {}, expected_params)
+        self.mock_handler.activate()
+        self.handler.invoke(LambdaName.DRIVER, {})
```
parthshahva: This looks to be causing issues in the travis test build. Are you sure its mocked or working as you intended here?

calvinnhieu: Environment variable issue

tests/unit/common/test_lambda_handler.py
```diff
@@ -0,0 +1,23 @@
+import unittest
+
+from botocore.stub import Stubber
```
parthshahva: Why are you using stubber instead of https://github.com/spulec/moto specific to test lambda invocation? In my opinion its a better library to use for testing functionality that involve calls to aws apis. It's more specific than just a general response stub

calvinnhieu: Agreed! Looks like with `Stubber` we're limited to verifying the function inputs and outputs, but not the actual state in AWS. Didn't know about `moto` but sgtm to use that instead

PR #53, opened 2018-10-05
Create dynamo handler class

calvinnhieu:
New class: `DyanmoHandler` is responsible for handling all interactions with DynamoDB tables. The current implementation only supports the initialization of the State table for tracking the execution state of filter merge jobs. 

New enum: `StateTableField` captures the field names of the State table in Dynamo

Alternatives:
Parallel state execution could be handled by a Step Functions implementation, however, a filter merge job requires support for a dynamic number of parallel executions which Step Functions currently [does not support](https://forums.aws.amazon.com/thread.jspa?threadID=244196). Until we find a better solution, we will keep the Dynamo implementation.


calvinnhieu:
TODO 10/5: write tests

tests/unit/common/test_dynamo_handler.py
```diff
@@ -0,0 +1,35 @@
...
+                StateTableField.COMPLETED_WORKER_EXECUTIONS.value: 0,
+                StateTableField.EXPECTED_MAPPER_EXECUTIONS.value: num_bundles,
+                StateTableField.COMPLETED_MAPPER_EXECUTIONS.value: 0,
+                StateTableField.EXPECTED_REDUCER_EXECUTIONS.value: 1,
+                StateTableField.COMPLETED_REDUCER_EXECUTIONS.value: 0
+            }
+        }
+        mock_client.add_response("put_item", {}, expected_params)
+        mock_client.activate()
+        self.handler.init_state_table(request_id, num_bundles)
```
parthshahva: This test doesn't have an input/output contract. What are you testing? Can we assert anything after this function has executed?

tests/unit/common/test_dynamo_handler.py
```diff
@@ -0,0 +1,35 @@
+import os
+import unittest
+
+from botocore.stub import Stubber
```
parthshahva: Why are you using stubber versus https://github.com/spulec/moto for dynamodb?

tests/unit/common/test_dynamo_handler.py
calvinnhieu: `init_state_table` eventually calls `put_item` on the Table resource. This test verifies the input params to `put_item`. OK to switch to `moto` to enable output verification

tests/unit/common/test_dynamo_handler.py
calvinnhieu: Didn't know about `moto`. Having the ability to verify output in AWS services sounds great - opting for `moto`

tests/__init__.py
```diff
@@ -1,6 +0,0 @@
-import os
```
calvinnhieu: Moved to tests/unit/__init__.py

PR #54, opened 2018-10-05
Port driver implementation from PoC

calvinnhieu:
_Ported driver implementation from [here](https://github.com/HumanCellAtlas/table-testing/tree/master/distributed_test/filter_merge_lambda)._

Driver is the first subtask of a filter merge job. The current implementation accepts a list of bundle fqids (fully qualified id; \<uuid\>.\<version\>), verifies correct parameter usage (`bundle_fqids` or `bundle_fqid_urls`, but not both), and invokes a `Mapper` subtask for each `bundle_fqid`.

### DIFF
--- a/daemons/filter_merge/app.py
+++ b/daemons/filter_merge/app.py
@@ -11,14 +11,17 @@ worker  - Apply user-defined filter query on chunk, write partial results to S3
 reducer - Combine partial results into final .zarr file in S3, return S3 location
 """
 
-from matrix.lambdas.filter_merge.driver import driver
+from matrix.lambdas.filter_merge.driver import Driver
 from matrix.lambdas.filter_merge.mapper import mapper
 from matrix.lambdas.filter_merge.worker import worker
 from matrix.lambdas.filter_merge.reducer import reducer
 
 
-def driver_handler():
-    driver()
+def driver_handler(event, context):
+    # TODO: better error handling
+    assert 'request_id' in event and 'bundle_fqids' in event and 'format' in event
+    driver = Driver()
+    driver.run(event['request_id'], event['bundle_fqids'], event['format'])
 
 
 def mapper_handler():
--- a/matrix/common/dynamo_handler.py
+++ b/matrix/common/dynamo_handler.py
@@ -17,15 +17,24 @@ class StateTableField(Enum):
     COMPLETED_REDUCER_EXECUTIONS = "CompletedReducerExecutions"
 
 
+class OutputTableField(Enum):
+    """
+    Field names for Output table in DynamoDB.
+    """
+    REQUEST_ID = "RequestId"
+    ROW_COUNT = "RowCount"
+
+
 class DynamoHandler:
     """
     Interface for interacting with DynamoDB Tables.
     """
     def __init__(self):
         self._dynamo = boto3.resource("dynamodb", region_name=os.environ['AWS_DEFAULT_REGION'])
-        self._state_table = self._dynamo.Table(os.environ['DYNAMO_STATE_TABLE_NAME'])
+        self._state_table = self._dynamo.Table(os.getenv("DYNAMO_STATE_TABLE_NAME"))
+        self._output_table = self._dynamo.Table(os.getenv("DYNAMO_OUTPUT_TABLE_NAME"))
 
-    def put_state_item(self, request_id, num_bundles):
+    def create_state_table_entry(self, request_id: str, num_bundles: int):
         """
         Put a new item in the DynamoDB table responsible for tracking task execution states and
         counts for a specified job.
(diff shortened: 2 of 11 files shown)


calvinnhieu:
TODO: write tests

parthshahva:
Will review once tests are added

daemons/filter_merge/app.py
```diff
@@ -11,14 +11,17 @@
...
+from matrix.lambdas.filter_merge.driver import Driver
 from matrix.lambdas.filter_merge.mapper import mapper
 from matrix.lambdas.filter_merge.worker import worker
 from matrix.lambdas.filter_merge.reducer import reducer
 
 
-def driver_handler():
-    driver()
+def driver_handler(event, context):
+    # TODO: better error handling
```
calvinnhieu: Do we want to start talking about logging/correlation id?

parthshahva: It will be easier to add in standardized logging/correlation id one PR rather than scattering the approach in multiple PRS. Lets focus on that after a passing functional test

matrix/common/dynamo_handler.py
```diff
@@ -17,17 +17,26 @@ class StateTableField(Enum):
...
     Interface for interacting with DynamoDB Tables.
     """
     def __init__(self):
         self._dynamo = boto3.resource("dynamodb", region_name=os.environ['AWS_DEFAULT_REGION'])
-        self._state_table = self._dynamo.Table(os.environ['DYNAMO_STATE_TABLE_NAME'])
+        self._state_table = self._dynamo.Table(os.getenv("DYNAMO_STATE_TABLE_NAME"))
+        self._output_table = self._dynamo.Table(os.getenv("DYNAMO_OUTPUT_TABLE_NAME"))
 
-    def init_state_table(self, request_id, num_bundles):
+    def put_state_item(self, request_id: str, num_bundles: int):
```
calvinnhieu: Is there any reason we would need to pass the `uuid` around instead of just the string?

parthshahva: string is fine

matrix/lambdas/api/core.py
```diff
@@ -1,23 +1,61 @@
...
-    LAMBDA_CLIENT.invoke(
-        FunctionName=os.environ['DRIVER_FN_NAME'],
-        InvocationType="Event",
-        Payload=json.dumps(body).encode(),
-    )
-    return "post_matrix", requests.codes.ok
+    format = body['format'] if 'format' in body else MatrixFormat.ZARR.value
 
+    # Validate input parameters
+    if not (has_ids or has_url):
```
parthshahva: combo of not and or leads to high code complexity

matrix/lambdas/api/core.py
```diff
@@ -1,23 +1,61 @@
...
+        'request_id': request_id,
+        'bundle_fqids': bundle_fqids,
+        'format': format,
+    }
+    lambda_handler = LambdaHandler()
+    lambda_handler.invoke(LambdaName.DRIVER, driver_payload)
+
+    return {'request_id': request_id,
+            'status': MatrixRequestStatus.IN_PROGRESS.value,
+            'key': "",
```
parthshahva: Lets investigate why these are included in the swagger object definition. Maybe split out the same response object currently shared between endpoints

matrix/lambdas/filter_merge/driver.py
```diff
@@ -1,2 +1,34 @@
...
+    def run(self, request_id: str, bundle_fqids: typing.List[str], format: str):
+        """
+        Initialize a filter merge job and spawn a mapper task for each bundle_fqid.
+
+        :param request_id: Filter merge job request ID
+        :param bundle_fqids: List of bundle fqids to be queried on
+        :param format: MatrixFormat file format of expression matrices
+        """
+        self.dynamo_handler.put_state_item(request_id, len(bundle_fqids))
+        self.dynamo_handler.put_output_item(request_id)
```
parthshahva: (nit) any better names here for put_state_item and put_output_item?

matrix/lambdas/api/core.py
calvinnhieu: Will go with `not has_ids and not has_url`

matrix/lambdas/api/core.py
calvinnhieu: I agree they should be different response objects/the current one should be updated. I'll ping @mckinsel to get some context (orange box is working against this schema right now?)

matrix/lambdas/filter_merge/driver.py
calvinnhieu: How's `create_state_table_entry` and `create_output_table_entry`?

PR #55, opened 2018-10-05
Rename aws resources

calvinnhieu:
Essentially, renaming Driver Lambda resources to reflect the decision to create 1 Lambda resource per subtask (driver, mapper, worker, reducer) in favor of 1 Lambda to handle all 4 subtasks. 

IAM role: `matrix-service-driver-daemon-${var.deployment_stage}` (per subtask)
IAM policy: `matrix-service-driver-daemon-${var.deployment_stage}` (per subtask)
S3 bucket for deployed zip: `matrix_service_lambda_deployment_bucket` (1 for all)
Lambda: `dcp-matrix-service-driver-${var.deployment_stage}` (per subtask)

The caveat is that a single S3 bucket for deployed zips will be shared amongst all 4 subtasks


terraform/modules/matrix-service/driver_lambda.tf
```diff
@@ -0,0 +1,85 @@
...
+  name = "matrix-service-driver-daemon-${var.deployment_stage}"
+  role = "${aws_iam_role.matrix_service_driver_lambda.name}"
+  policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "LogsPolicy",
+      "Effect": "Allow",
+      "Resource": "arn:aws:logs:*:*:*",
```
parthshahva: Please lock this resource path down to only this lambda's cloudwatch log group

terraform/modules/matrix-service/driver_lambda.tf
```diff
@@ -0,0 +1,85 @@
...
+        "logs:PutLogEvents"
+      ]
+    },
+    {
+      "Sid": "DynamoPolicy",
+      "Effect": "Allow",
+      "Action": [
+        "dynamodb:PutItem"
+      ],
+      "Resource": "arn:aws:dynamodb:us-east-1:861229788715:table/dcp-matrix-service-state-table-${var.deployment_stage}"
```
parthshahva: please add account id to your environment variables + terraform variables and interpolate. Any hard coded account ids will make it difficult to deploy to another account (ie prod)

terraform/modules/matrix-service/driver_lambda.tf
```diff
@@ -0,0 +1,85 @@
...
+    },
+    {
+      "Sid": "LambdaPolicy",
+      "Effect": "Allow",
+      "Action": [
+        "lambda:InvokeFunction"
+      ],
+      "Resource": "arn:aws:lambda:us-east-1:861229788715:function:dcp-matrix-service-mapper-${var.deployment_stage}"
```
parthshahva: account id interpolation required as mentioned above

terraform/modules/matrix-service/driver_lambda.tf
```diff
@@ -0,0 +1,85 @@
...
+  s3_key           = "driver_daemon.zip"
+  role             = "${aws_iam_role.matrix_service_driver_lambda.arn}"
+  handler          = "app.driver_handler"
+  runtime          = "python3.6"
+  timeout          = 300
+
+  environment {
+    variables = {
+        DEPLOYMENT_STAGE = "${var.deployment_stage}"
+        LAMBDA_DRIVER_FUNCTION_NAME="dcp-matrix-service-driver-${var.deployment_stage}"
```
parthshahva: why does the driver need all this lambda variables. Doesn't it only need to know about the next one in the sequence?

calvinnhieu: Hm yea that makes sense. Problem is importing https://github.com/HumanCellAtlas/matrix-service/pull/52/files#diff-6d63faf8105e81f612d9bfee6d1d5aefR8 requires all 4 to be set. I personally think enums are a nice plus for programmatically organizing keys. Thoughts?

parthshahva: Ah I see. I agree the enum is nice. You could just change the code in the link to utilize os.getenv. That way any variables not appropriate for a lambda will get set to None and fail on invocation attempt of a lambda it should not have permission to invoke.

terraform/modules/matrix-service/driver_lambda.tf
calvinnhieu: Done

PR #57, opened 2018-10-05
Tf restructure and tf for worker

parthshahva:
1) TF for dcp-matrix-service-worker
2) Restructure of terraform + shared deployment s3 bucket between lambda tf


terraform/modules/matrix-service/lambdas/worker_lambda.tf
```diff
@@ -0,0 +1,82 @@
...
+        "lambda:InvokeFunction"
+      ],
+      "Resource": "arn:aws:lambda:${var.aws_region}:${var.account_id}:function:dcp-matrix-service-reducer-${var.deployment_stage}"
+    }
+  ]
+}
+EOF
+}
+
+resource "aws_s3_bucket" "matrix_service_lambda_deployment_bucket" {
```
calvinnhieu: Remove

PR #60, opened 2018-10-09
Hca store

parthshahva:
Heavily borrowed from https://github.com/HumanCellAtlas/hca-matrix-utils. Importing in hca store, util functions, and tests

### DIFF
--- a/matrix/common/hca_store.py
+++ b/matrix/common/hca_store.py
@@ -1,0 +1,129 @@
+# -*- coding: utf-8 -*-
+"""Classes for working with expression matrices serialized in the DCP."""
+
+from collections import MutableMapping
+import hashlib
+
+
+class HCAStore(MutableMapping):
+    """
+    Zarr compatible interface to expression matrices stored in a DCP analysis bundle.
+    Parameters
+    ----------
+    dss_client : hca.dss.DSSClient
+        Client used to access the DCP Data Storage System.
+    bundle_uuid : str
+        DCP uuid of the analysis bundle.
+    bundle_version : str, optional
+        DCP version of the analysis bundle. (default is None, which means use
+        the latest version of the bundle)
+    replica : str, optional
+        Replica of the DCP from which to access data. (default is "aws")
+    Attributes
+    ----------
+    bundle_uuid
+    bundle_version
+    Methods
+    -------
+    Provides the MutableMapping interface, as required by zarr. But bad news: the DCP isn't
+    mutable, so mutations raise NotImplementedError with a helpful note.
+    """
+
+    # The DSS doesn't allow "directories", which really means it doesn't allow
+    # file keys with a '/' in the name since none of the storage backends has a
+    # directory concept anyway. So, we replace '/' with a different separator.
+    _separator_char = "!"
+
+    # There are possibly many files in the analysis bundle, but we only want to
+    # read those that are associated with the expression matrix. We can
+    # identify those via a prefix in the file name.
+    _zarr_file_prefix = "expression_matrix"
+
+    def _transform_key(self, key):
+        return key.replace("/", self._separator_char)
+
+    def __init__(self, dss_client, bundle_uuid, bundle_version=None, replica="aws"):
+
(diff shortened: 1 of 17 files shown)


matrix/common/util.py
```diff
@@ -0,0 +1,42 @@
+import pandas
+import hca
+import zarr
+
+
+def get_client(dss_instance=None):
```
calvinnhieu: What do you think of having this as a @property in `HCAStore`? I think it comes down to, do we know if other parts of matrix service are going to need to access the DSSClient outside of `HCAStore`?

requirements.txt
```diff
@@ -6,3 +6,7 @@ itsdangerous==0.24
 MarkupSafe==1.0
 PyYAML==3.13
 requests==2.19.1
+numpy==1.15.2
+pytest==3.8.2
```
calvinnhieu: requirements-dev.txt?

tests/unit/__init__.py
```diff
@@ -0,0 +1,27 @@
+test_bundle_spec = {
```
calvinnhieu: I think this is fine, but also that storing this in a file in `matrix/tests/res/` would be cleaner. Fine either way

tests/unit/common/test_hca_store.py
```diff
@@ -0,0 +1,38 @@
...
+import pytest
+import zarr
+
+from matrix.common.hca_store import HCAStore
+from .. import test_bundle_spec
+
+TEST_DSS_HOST = 'https://dss.integration.data.humancellatlas.org/v1'
+
+
+class TestLambdaHandler(unittest.TestCase):
```
calvinnhieu: Name

tests/unit/common/test_util.py
```diff
@@ -0,0 +1,26 @@
+import numpy
+import unittest
+import pytest
+
+from matrix.common.util import get_client, convert_zarr_store_to_pandas_df
+from matrix.common.hca_store import HCAStore
+from .. import test_bundle_spec
+
+
+class TestUtils(unittest.TestCase):
```
calvinnhieu: nit: `TestUtil` to match file name

matrix/common/util.py
```diff
@@ -0,0 +1,42 @@
...
+    client = hca.dss.DSSClient()
+    if not dss_instance:
+        client.host = "https://dss.integration.data.humancellatlas.org/v1"
+    else:
+        client.host = "https://dss.{dss_instance}.data.humancellatlas.org/v1".format(
+            dss_instance=dss_instance)
+    return client
+
+
+def convert_zarr_store_to_pandas_df(zarr_store, concat=False):
```
calvinnhieu: Function signature types, + others

tests/unit/common/test_hca_store.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test_hca_store_read(self):
```
calvinnhieu: Could use more failure case tests, but would require better error handling. I'm not sure we understand this code path well enough to implement all of those. I'm OK with adding to these tests once we have v0 running.

matrix/common/hca_store.py
```diff
@@ -0,0 +1,129 @@
+# -*- coding: utf-8 -*-
+"""Classes for working with expression matrices serialized in the DCP."""
+
+from collections import MutableMapping
+import hashlib
+
+
+class HCAStore(MutableMapping):
```
calvinnhieu: What do you think of `HCAZarrStore` to communicate that it's zarr compatible?

parthshahva:
Thanks for the comments @calvinnhieu. You did the dirty work of digging through the copy and paste job. I was thinking we could clean it up after getting Marcus' initial implementation merged in but now agree with all your comments and will push up before merging.

calvinnhieu:
Sounds good!

matrix/common/hca_store.py
calvinnhieu: Prefer `DSSZarrStore`

PR #62, opened 2018-10-09
refactor out filter_merge references

calvinnhieu:
In favor of individual lambdas for each "filter merge" stage (driver, mapper, worker, reducer)


daemons/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""
```
calvinnhieu: moved existing comment

calvinnhieu:
Tested with driver deploy (`cd daemons && make deploy`) and successful API calls via local API server.

PR #63, opened 2018-10-09
Create Mapper lambda

calvinnhieu:
Probably best that our terraform updates are all in develop


terraform/modules/matrix-service/lambdas/driver_lambda.tf
```diff
@@ -44,7 +44,10 @@ resource "aws_iam_role_policy" "matrix_service_driver_lambda" {
       "Action": [
         "dynamodb:PutItem"
       ],
-      "Resource": "arn:aws:dynamodb:${var.aws_region}:${var.account_id}:table/dcp-matrix-service-state-table-${var.deployment_stage}"
+      "Resource": [
```
calvinnhieu: This is actually a bug fix for driver... Including it with the other tf changes

PR #65, opened 2018-10-11
refactor of utils to pandas utils

parthshahva:
1) Refactored utils.py to pandas_utils.py
2) Introduced pass_filter_string_through_matrix_pandas_dfs and corresponding test


matrix/common/pandas_utils.py
```diff
@@ -0,0 +1,41 @@
+import pandas
+from pandas.core.frame import DataFrame
+from zarr import Group
+
+
+def convert_dss_zarr_root_to_subset_pandas_dfs(zarr_root: Group, start_row: int, end_row: int):
```
calvinnhieu: thoughts on breaking this into `get_expression_dataframe` and `get_qc_dataframe`?

calvinnhieu: Or - does this currently return a Tuple or just 2 values? Wouldn't be opposed to a Tuple if both dfs are tightly coupled

matrix/common/pandas_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    expression_df = pandas.DataFrame(data=zarr_root.expression[start_row:end_row],
+                                     index=zarr_root.cell_id[start_row:end_row],
+                                     columns=zarr_root.gene_id)
+    qc_df = pandas.DataFrame(data=zarr_root.cell_metadata[start_row:end_row],
+                             index=zarr_root.cell_id[start_row:end_row],
+                             columns=zarr_root.cell_metadata_name[start_row:end_row])
+    return expression_df, qc_df
+
+
+def pass_filter_through_pandas_dfs(filter_query: str, exp_df: DataFrame, qc_df: DataFrame):
```
calvinnhieu: Thoughts on renaming to `apply_filter`? And if above fn returns a Tuple, this could accept a Tuple too

matrix/common/pandas_utils.py
parthshahva: Lets keep this as one for now. As of now, there is no functionality in the matrix service that doesn't interact with both dataframes simultaneously

PR #66, opened 2018-10-11
Implement mapper daemon

calvinnhieu:


calvinnhieu:
TODO: fix errors when lambda is deployed

PR #67, opened 2018-10-11
empty shell for worker lambda deployment

parthshahva:
Deployment instructions, directory shell, and empty methods for worker daemon

### DIFF
--- a/daemons/worker/app.py
+++ b/daemons/worker/app.py
@@ -1,0 +1,12 @@
+from matrix.lambdas.daemons.worker import Worker
+
+
+def worker_handler(event, context):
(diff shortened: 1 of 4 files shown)


daemons/worker/app.py
```diff
@@ -0,0 +1,12 @@
+from matrix.lambdas.daemons.worker import Worker
+
+
+def worker_handler(event, context):
+    assert 'request_id' in event and 'work_chunk_spec' in event and 'format' in event and 'filter_string' in event
```
calvinnhieu: using `worker_chunk_spec` instead

PR #68, opened 2018-10-11
shell of class for zarr to s3 writer

parthshahva:
Shell of ZarrToS3Writer class with only initial method in run() implemented

### DIFF
--- a/matrix/common/zarr_to_s3_writer.py
+++ b/matrix/common/zarr_to_s3_writer.py
@@ -25,7 +25,7 @@ class ZarrToS3Writer():
         }
         self.order = "C"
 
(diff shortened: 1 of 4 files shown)


matrix/common/zarr_to_s3_writer.py
```diff
@@ -0,0 +1,54 @@
...
+        self.cells_per_chunk = cells_per_chunk
+        self.compressor = compressor
+        self.dtypes = {
+            "data": "<f4",
+            "cell_name": "<U64",
+            "qc_values": "<f4"
+        }
+        self.order = "C"
+
+    def run(self, request_id: str, filtered_exp_df: DataFrame, filtered_qc_df: DataFrame):
```
calvinnhieu: `run` -> `write`?

PR #71, opened 2018-10-12
Implement basic GET endpoint

calvinnhieu:
GET endpoint checks DyanmoDB for `CompletedReducerExecutions` to determine if result is ready. If ready, return `200` and the S3 prefix key pointing to where the request results are. If not ready, also return `200`(?) with appropriate messaging.

### DIFF
--- a/config/iam_policy_templates/matrix-service-api-lambda.json
+++ b/config/iam_policy_templates/matrix-service-api-lambda.json
@@ -15,7 +15,14 @@
       "Action": [
         "lambda:InvokeFunction"
       ],
-      "Resource": "arn:aws:lambda:us-east-1:861229788715:function:dcp-matrix-service-driver-dev"
+      "Resource": "arn:aws:lambda:us-east-1:${account_id}:function:dcp-matrix-service-driver-${DEPLOYMENT_STAGE}"
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "dynamodb:GetItem"
+      ],
+      "Resource": "arn:aws:dynamodb:us-east-1:${account_id}:table/dcp-matrix-service-state-table-${DEPLOYMENT_STAGE}"
     }
   ]
 }
--- a/config/iam_policy_templates/matrix-service-api-lambda.json
+++ b/config/iam_policy_templates/matrix-service-api-lambda.json
@@ -16,17 +16,6 @@
         "lambda:InvokeFunction"
       ],
       "Resource": "arn:aws:lambda:us-east-1:861229788715:function:dcp-matrix-service-driver-dev"
-    },
-    {
-      "Effect": "Allow",
-      "Action": [
-        "s3:ListBucket",
-        "s3:GetObject"
-      ],
-      "Resource": [
-        "arn:aws:s3:::dcp-matrix-service-results-bucket-dev",
-        "arn:aws:s3:::dcp-matrix-service-results-bucket-dev:*"
-      ]
     }
   ]
 }
--- a/config/iam_policy_templates/matrix-service-api-lambda.json
+++ b/config/iam_policy_templates/matrix-service-api-lambda.json
@@ -16,6 +16,17 @@
         "lambda:InvokeFunction"
       ],
       "Resource": "arn:aws:lambda:us-east-1:861229788715:function:dcp-matrix-service-driver-dev"
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "s3:ListBucket",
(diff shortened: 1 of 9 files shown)


config/iam_policy_templates/matrix-service-api-lambda.json
```diff
@@ -16,6 +16,17 @@
         "lambda:InvokeFunction"
       ],
       "Resource": "arn:aws:lambda:us-east-1:861229788715:function:dcp-matrix-service-driver-dev"
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "s3:ListBucket",
```
parthshahva: It's a public read bucket. Why is this policy needed?

matrix/lambdas/api/core.py
```diff
@@ -55,11 +59,31 @@ def post_matrix(body: dict):
             }, requests.codes.accepted
 
 
+# TODO: write tests
 def get_matrix(request_id: str):
+    dynamo_handler = DynamoHandler()
+    # TODO: error handling: invalid request_id
+    ready, ready = dynamo_handler.increment_table_field(DynamoTable.STATE_TABLE,
```
parthshahva: Why are you incrementing the reducer executions in the get_matrix request?

matrix/lambdas/api/core.py
```diff
@@ -55,11 +59,31 @@ def post_matrix(body: dict):
...
-            'status': "In Progress",
-            'key': "sample key",
-            'eta': "sample eta",
-            'message': "sample message",
+            'status': "In progress",
+            'key': "N/A",
+            'eta': "N/A",
+            'message': "Request {request_id} is still processing. Please try again later.",
             'links': [],
             }, requests.codes.ok
```
parthshahva: is there a better http status code we can use here?

matrix/lambdas/api/core.py
calvinnhieu: Hacky way to read from the table copied from port over. Will rewrite

config/iam_policy_templates/matrix-service-api-lambda.json
calvinnhieu: Good call. Removing

matrix/lambdas/api/core.py
calvinnhieu: Possibly 303 See Other?
https://github.com/requests/requests/blob/master/requests/status_codes.py

brianraymor: **Pick your adventure** There is no standard.

Even Microsoft suggests both of the common patterns:

[Azure REST API](https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design#asynchronous-operations) recommends:
* a 202 response to the POST with the monitor in the Location: header
* a 303 response to the GET when the operation is successful and the created resource URI in the Location:header

[Microsoft REST API Guidelines](https://github.com/Microsoft/api-guidelines/blob/vNext/Guidelines.md#1327-the-typical-flow-polling) recommends:
* a 202 response to the POST with the monitor in a proprietary header
* a 200 response to the GET when the operation is successful with the created resource URI in the message body (_resourceLocation_)

brianraymor: @mckinsel - If we continue to embed the created Resource URI in the message body, could we use a _self-documenting_ name (based on Dave Roger's questions). If _key_ was renamed _expression-matrix_ (or _matrix-location_) wouldn't that be relatively clear?

calvinnhieu:
Thanks for the comments @brianraymor - I agree we can just go with one and my vote is for the 303 with location header. I've reached out to @hannes-ucsc @NoopDog for input on the API response changes. I don't have an issue with removing `links` in v0 - @mckinsel ?

matrix/lambdas/api/core.py
```diff
@@ -46,20 +53,44 @@ def post_matrix(body: dict):
...
-            'key': "sample key",
-            'eta': "sample eta",
-            'message': "sample message",
-            'links': [],
-            }, requests.codes.ok
+    try:
+        dynamo_handler = DynamoHandler()
+        job_state = dynamo_handler.get_table_item(DynamoTable.STATE_TABLE, request_id)
+    except MatrixException as ex:
+        return ConnexionResponse(status_code=ex.status,
```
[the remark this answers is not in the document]
calvinnhieu: In this specific case, I'm providing a more user friendly/context specific error message than the one set on the exception. Also, connexion is validating the response schema as defined in the swagger spec

PR #73, opened 2018-10-15
expose DSSZarrStore data as properties

calvinnhieu:


matrix/common/utils.py
```diff
@@ -0,0 +1,30 @@
+import pandas
```
parthshahva: why is this in the PR? I donn't see any file named utils.py anymore. It should be pandas_utils.py. We are also planning on moving this function into the dss zarr store class?

calvinnhieu: Looks like my changes got mixed up. Agreed this shouldn't be here. Removing from `develop`

PR #75, opened 2018-10-16
WIP worker

parthshahva:
Things left
1) more tests
2) code cleanup


daemons/worker/Makefile
```diff
@@ -1,4 +1,4 @@
-include ../../common.mk
+# include ../../common.mk
```
calvinnhieu: Uncomment

daemons/worker/app.py
```diff
@@ -2,11 +2,12 @@
...
+    assert 'request_id' in event and 'worker_chunk_spec' in event and 'format' in event
 
     worker_chunk_spec = event['worker_chunk_spec']
     assert 'bundle_uuid' in worker_chunk_spec and 'bundle_version' in worker_chunk_spec
     assert 'start_row' in worker_chunk_spec and 'num_rows' in worker_chunk_spec
 
-    worker = Worker()
-    worker.run(event['request_id'], event['filter_string'], worker_chunk_spec)
+    worker = Worker(event["request_id"])
+    worker.run(event["format"], event["worker_chunk_spec"])
```
calvinnhieu: single quotes for key access

matrix/common/dynamo_handler.py
```diff
@@ -44,6 +44,20 @@ def __init__(self):
         self._state_table = self._dynamo.Table(DynamoTable.STATE_TABLE.value)
         self._output_table = self._dynamo.Table(DynamoTable.OUTPUT_TABLE.value)
 
+    def _get_dynamo_table_from_name(self, dynamo_table_name: str):
```
calvinnhieu: Personally think name is verbose for this impl, but would be appropriate if `_get_dynamo_table...` was overloaded. Would opt for `_get_dynamo_table`

calvinnhieu: `dynamo_table_name` should use enum and function impl accesses enum `.value`

matrix/common/dynamo_handler.py
```diff
@@ -76,22 +91,34 @@ def create_output_table_entry(self, request_id: str):
...
+        """Retrieves dynamobdb item corresponding with request_id in specified table
+        Args:
+            table: DynamoTable enum
+            request_id: request id key in table
+        Returns:
+            item: dynamodb item
+        """
+        dynamo_table = self._get_dynamo_table_from_name(table.name)
+        item = dynamo_table.get_item(
+            Key={"RequestId": request_id},
```
calvinnhieu: single quotes for keys

matrix/common/dynamo_handler.py
calvinnhieu: Oh I see what you're doing now. I'm fine with this, but I think using the enum object the whole time might be cleaner

matrix/common/dynamo_handler.py
```diff
@@ -44,6 +44,20 @@ def __init__(self):
...
+    def _get_dynamo_table_resource_from_enum(self, dynamo_table: DynamoTable):
+        """Retrieve dynamo table resource for a given dynamo table name.
+
+        Input:
+            dynamo_table: (DynamoTable), Enum
+        Output:
+            boto3 dynamodb resource
+        """
+        name = dynamo_table.name
+        if name == "STATE_TABLE":
```
calvinnhieu: `if dynamo_table == DynamoTable.STATE_TABLE`, thoughts?

matrix/common/s3_zarr_store.py
```diff
@@ -0,0 +1,188 @@
...
+    "order": "C"
+}
+
+
+class S3ZarrStore:
+
+    def __init__(self, request_id: str):
+        self._request_id = request_id
+        self._results_bucket = os.environ['S3_RESULTS_BUCKET']
+        self._zarr_lock_table_name = os.environ['DYNAMO_LOCK_TABLE_NAME']
```
calvinnhieu: I think it makes more sense for this to live in `Lock`

matrix/common/s3_zarr_store.py
```diff
@@ -0,0 +1,188 @@
...
+    def write_from_pandas_dfs(self, exp_df: DataFrame, qc_df: DataFrame, num_rows: int):
+        """Write specified number of rows from matrix dataframes to s3 results bucket.
+
+        Input:
+            exp_df: (DataFrame) expression pandas dataframe from dss zarr store
+            qc_df: (DataFrame) qc values pandas dataframe from dss zarr store
+            num_rows: (int) number of rows to write from input dataframes
+        """
+        self.exp_df = exp_df
+        self.qc_df = qc_df
```
calvinnhieu: Feels like a weird pattern. Would opt for one of the following:
- set on init
- setter method
- always pass as function params

matrix/common/s3_zarr_store.py
```diff
@@ -0,0 +1,188 @@
...
+            elif dset == "cell_metadata":
+                values = self.qc_df.values
+            elif dset == "cell_id":
+                values = self.exp_df.index.values
+            full_dest_key = f"s3://{self._results_bucket}/{self._request_id}.zarr/{dset}/{chunk_idx}.0"
+            print(f"Writing {dset} to {full_dest_key}")
+            if values.ndim == 2:
+                chunk_shape = (self._cells_per_chunk, values.shape[1])
+            else:
+                chunk_shape = (self._cells_per_chunk,)
```
calvinnhieu: Is this intentional?

matrix/lambdas/daemons/worker.py
```diff
@@ -1,21 +1,89 @@
...
+        """Process and write one chunk of dss bundle matrix to s3 and
+        invoke reducer lambda when last worker job is completed.
 
-    def run(self, request_id: str, filter_string: str, worker_chunk_spec: dict):
+        Inputs:
+        format: (str) Expected file format of request
+        worker_chunk_spec: (dict) Information about input bundle and matrix row indices to process
         """
-        Filter one work chunk and invoke reducer lambda when last worker job is completed.
+        self._parse_worker_chunk_spec(worker_chunk_spec)
```
calvinnhieu: I'm wondering if we're doing this if we should just pass in the parameters flat.

matrix/lambdas/daemons/worker.py
```diff
@@ -1,21 +1,89 @@
...
+        dss_zarr_store = DSSZarrStore(bundle_uuid=self._bundle_uuid,
+                                      bundle_version=self._bundle_version,
+                                      dss_instance=self._deployment_stage)
+        group = zarr.group(store=dss_zarr_store)
+        exp_df, qc_df = convert_dss_zarr_root_to_subset_pandas_dfs(group, self._input_start_row, self._input_end_row)
+        s3_zarr_store = S3ZarrStore(self._request_id)
+        s3_zarr_store.write_from_pandas_dfs(exp_df, qc_df, self._num_rows)
+
+        field_value = StateTableField.COMPLETED_WORKER_EXECUTIONS.value
+        self.dynamo_handler.increment_table_field(DynamoTable.STATE_TABLE, self._request_id, field_value, 1)
```
calvinnhieu: I think this function should accept enums as params

matrix/lambdas/daemons/worker.py
```diff
@@ -1,21 +1,89 @@
...
+            worker_chunk_spec: (dict) keys expected 'bundle_uuid', 'bundle_version',
+                               'start_row', 'num_rows'
+        """
+        self._bundle_uuid = worker_chunk_spec['bundle_uuid']
+        self._bundle_version = worker_chunk_spec['bundle_version']
+        self._input_start_row = worker_chunk_spec['start_row']
+        self._num_rows = worker_chunk_spec['num_rows']
+        self._input_end_row = self._input_start_row + self._num_rows
+
+    def _check_if_all_workers_and_mappers_for_request_are_complete(self, request_id):
```
calvinnhieu: Prefer simpler function names. I don't think function names need to explain what the function is doing. Suggestions:
- `_is_reducer_ready`
- `_is_last_worker`

tests/unit/lambdas/daemons/test_worker.py
```diff
@@ -0,0 +1,93 @@
...
+        mock_zarr_group.return_value = None
+        mock_df_conversion.return_value = (None, None)
+        mock_is_complete.return_value = True
+        self.worker.run(self.format_string, self.worker_chunk_spec)
+        expected_calls = [
+            mock.call(LambdaName.REDUCER, {'request_id': self.request_id,
+                                           'format': self.format_string}),
+        ]
+        self.assertEqual(mock_lambda_handler_invoke.call_count, 1)
+        mock_lambda_handler_invoke.assert_has_calls(expected_calls)
```
calvinnhieu: For asserting single calls, this works too:
`mock_lambda_handler.assert_called_once_with(LambdaName.REDUCER, {'request_id': self.request_id, 'format': self.format_string})`

matrix/common/s3_zarr_store.py
```diff
@@ -0,0 +1,188 @@
...
+            output_row_end = int(output_row_start + input_row_end - input_row_start)
+            print(f"Writing {input_row_start}:{input_row_end} --> {output_row_start}:{output_row_end}")
+
+            input_bounds = (input_row_start, input_row_end)
+            output_bounds = (output_row_start, output_row_end)
+            self._write_row_data_to_results_chunk(chunk_idx, input_bounds, output_bounds)
+            row_count = input_row_end - input_row_start
+            written_rows += row_count
+
+    def _write_row_data_to_results_chunk(self, chunk_idx: int, input_bounds: tuple, output_bounds: tuple):
```
calvinnhieu: Is this tested?

matrix/common/s3_zarr_store.py
parthshahva: Asked Marcus. This does seem weird. Missed on the first pass

matrix/lambdas/daemons/worker.py
parthshahva: Leave as is for now but good to note for future improvement.

parthshahva: Leavig a comment in code to track this

matrix/common/s3_zarr_store.py
parthshahva: Left comment here to add test here.  Lower priority for this week. Codecov tool for all future PRs is a great idea so that we keep marching in the right direction

PR #77, opened 2018-10-16
rename reducer vars

parthshahva:
Rename reduce vars to reflect naming convention for access to properties in dss zarr store group


calvinnhieu:
Is this tested? The current implementation works and the reducer should be reading from S3 zarr store (worker outputs).

parthshahva:
@calvinnhieu I changed the outputs from the worker to match dss zarr store property names. This PR matches the switch in names

PR #80, opened 2018-10-18
terraform for batch converter

parthshahva:
1) Vpc terraform setup
2) Converter batch terraform setup
3) Variable syntax cleanup
4) Addition to terraform Makefiles of import commands


terraform/modules/matrix-service/infra/converter_batch.tf
```diff
@@ -0,0 +1,188 @@
...
+  name = "dcp-matrix-converter-job-${var.deployment_stage}"
+  policy = <<POLICY
+{
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Sid": "s3",
+        "Effect": "Allow",
+        "Action": [
+          "s3:*"
```
calvinnhieu: Do we want to restrict this?

calvinnhieu:
Do you want to PR this into `develop`?

terraform/modules/matrix-service/infra/converter_batch.tf
mckinsel: I think it should just need `s3:PutObject` if `s3:ListBucket` and `s3:GetObject` are open to everyone.

terraform/modules/matrix-service/infra/converter_batch.tf
```diff
@@ -0,0 +1,221 @@
...
+  policy_arn = "${aws_iam_policy.converter_job_policy.arn}"
+}
+
+resource "aws_batch_job_definition" "converter_job_def" {
+    name = "dcp-matrix-converter-job-definition-${var.deployment_stage}"
+    type = "container"
+    container_properties = <<CONTAINER_PROPERTIES
+{
+    "command": [],
+    "image": "humancellatlas/matrix-converter:latest",
```
mckinsel: Maybe better if I tag an image?

parthshahva: @mckinsel I'm actually using slightly different code/image to get it to work with batch. That WIP PR is coming up shortly. I'll be deploying the image to humancellatlas/matrix-converter in dockerhub. I'm including a concept of VERSION. I agree. I'll replace latest with version here.

terraform/modules/matrix-service/infra/converter_batch.tf
parthshahva: sounds good

terraform/envs/staging/terraform.tfvars.example
```diff
@@ -1 +1,4 @@
 deployment_stage = ""
+account_id = ""
+aws_region = ""
+converter_cluster_ami_id=""
```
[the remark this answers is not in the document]
parthshahva: fixed

PR #83, opened 2018-10-18
Make matrix results bucket anonymously readable

mckinsel:


mckinsel:
I think you're right, but the docs say that `s3:GetObject` should be covered by that too. However, it wasn't working for me when I tried anonymous access.

PR #86, opened 2018-10-19
Fix issue with intermittent blosc failures

mckinsel:
See https://github.com/pangeo-data/pangeo/issues/196 for discussion of
the problem. Retry on RuntimeError from blosc.

### DIFF
--- a/matrix/common/s3_zarr_store.py
+++ b/matrix/common/s3_zarr_store.py
@@ -128,7 +128,6 @@ class S3ZarrStore:
                         num_tries += 1
                         delay *= 1.6
 
-
                 arr.setflags(write=1)
                 arr[output_bounds[0]:output_bounds[1]] = values[input_bounds[0]:input_bounds[1]]
                 self.s3_file_system.open(full_dest_key, 'wb').write(ZARR_OUTPUT_CONFIG['compressor'].encode(arr))
--- a/matrix/common/s3_zarr_store.py
+++ b/matrix/common/s3_zarr_store.py
@@ -1,6 +1,7 @@
 import os
 import math
 import json
+import time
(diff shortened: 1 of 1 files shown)


matrix/common/s3_zarr_store.py
```diff
@@ -128,7 +128,6 @@ def _write_row_data_to_results_chunk(self, chunk_idx: int, input_bounds: tuple,
                         num_tries += 1
                         delay *= 1.6
 
-
```
calvinnhieu: Lol

matrix/common/s3_zarr_store.py
```diff
@@ -102,15 +103,31 @@ def _write_row_data_to_results_chunk(self, chunk_idx: int, input_bounds: tuple,
...
-                            self.s3_file_system.open(full_dest_key, 'rb').read()),
-                        dtype=dtype).reshape(chunk_shape, order=ZARR_OUTPUT_CONFIG['order'])
-                except FileNotFoundError:
-                    arr = numpy.zeros(shape=chunk_shape,
-                                      dtype=dtype)
-                    print("Created new array")
+                # This is a graceless workaround for the issue identified here:
+                # https://github.com/pangeo-data/pangeo/issues/196
+                # Sometimes blosc.decode raises a RuntimeError that goes away on retry
+                # TODO: factor out the retry logic
```
calvinnhieu: I'm thinking this would look like a decorator that we could apply on a function that extracts the decomp logic? Fine for now

brianraymor: @mckinsel - would it make sense to also open a tracking issue on github in addition to the **TODO**?

PR #87, opened 2018-10-19
WIP of batching bundles to workers

mckinsel:


matrix/lambdas/daemons/driver.py
```diff
@@ -21,15 +24,20 @@ def run(self, request_id: str, bundle_fqids: typing.List[str], format: str):
...
         :param format: MatrixFormat file format of expression matrices
         """
-        self.dynamo_handler.create_state_table_entry(request_id, len(bundle_fqids))
+        num_expected_mappers = math.ceil(len(bundle_fqids) / self._bundles_per_worker)
+        self.dynamo_handler.create_state_table_entry(request_id, num_expected_mappers)
         self.dynamo_handler.create_output_table_entry(request_id)
 
-        for fqid in bundle_fqids:
-            bundle_uuid, bundle_version = fqid.split(".", 1)
+        iter_args = [iter(bundle_fqids)] * self._bundles_per_worker
```
calvinnhieu: Could this and `zip_longest` be pulled out to a separate function to be more descriptive such as `_group_bundle_fqids(bundle_fqids, self._bundles_per_worker)`

matrix/lambdas/daemons/driver.py
```diff
@@ -21,15 +24,20 @@ def run(self, request_id: str, bundle_fqids: typing.List[str], format: str):
...
+        for bundle_fqid_group in itertools.zip_longest(*iter_args):
+            bundle_fqid_group = list(filter(None, bundle_fqid_group))
+
+            bundle_uuids = [fqid.split(".", 1)[0] for fqid in bundle_fqid_group]
+            bundle_versions = [fqid.split(".", 1)[1] for fqid in bundle_fqid_group]
```
calvinnhieu: Think it makes more sense to pass in `bundle_fqids` and perform the splits in the distributed mappers. Not a big deal, just a preference

matrix/lambdas/daemons/mapper.py
```diff
@@ -68,7 +69,7 @@ def _get_worker_payload(self, worker_chunk_spec: dict) -> dict:
         }
 
     @staticmethod
-    def _get_chunk_specs(bundle_uuid: str, bundle_version: str) -> typing.List[dict]:
+    def _get_chunk_specs(bundle_uuids: typing.List[str], bundle_versions: typing.List[str]) -> typing.List[dict]:
```
calvinnhieu: Actually I did want to (and will) to refactor this to operate on a single bundle and run inside the for loop. Just noting

matrix/lambdas/daemons/mapper.py
```diff
@@ -77,16 +78,21 @@ def _get_chunk_specs(bundle_uuid: str, bundle_version: str) -> typing.List[dict]
...
+
+        chunk_work_spec = []
+        for bundle_uuid, bundle_version in zip(bundle_uuids, bundle_versions):
+            zarr_store = DSSZarrStore(bundle_uuid,
+                                      bundle_version=bundle_version,
+                                      dss_instance=os.environ['DEPLOYMENT_STAGE'])
+            root = zarr.group(store=zarr_store)
+
+            rows_per_chunk = root.expression.chunks[0]
+            total_chunks = root.expression.nchunks
```
calvinnhieu: `root.expression` is actually accessible directly from `zarr_store` now (my bad haven't updated): `zarr_store.expression`. Can remove dependency on `zarr` in this file

matrix/lambdas/daemons/mapper.py
```diff
@@ -46,8 +48,7 @@ def run(self, bundle_uuid: str, bundle_version: str):
         print(f"{self.request_id} Invoking {len(worker_chunk_specs)} worker lambdas...")
         for worker_chunk_spec in worker_chunk_specs:
             self.lambda_handler.invoke(LambdaName.WORKER, self._get_worker_payload(worker_chunk_spec))
-            print(f"{self.request_id} Worker invoked: start row "
-                  f"{worker_chunk_spec['start_row']}, num rows {worker_chunk_spec['num_rows']}")
+            print(f"{self.request_id} Worker invoked {worker_chunk_spec}")
```
calvinnhieu: Wait, it looks like this still invokes 1 Worker for every chunk. See *

matrix/lambdas/daemons/mapper.py
```diff
@@ -77,16 +78,21 @@ def _get_chunk_specs(bundle_uuid: str, bundle_version: str) -> typing.List[dict]
...
+
+            rows_per_chunk = root.expression.chunks[0]
+            total_chunks = root.expression.nchunks
+
+            chunk_work_spec.extend(
+                [{"bundle_uuid": bundle_uuid,
+                  "bundle_version": bundle_version,
+                  "start_row": n * rows_per_chunk,
+                  "num_rows": rows_per_chunk}
+                 for n in range(total_chunks)])
```
calvinnhieu: This still produces a flat array of worker specs, it just processes a list of bundles instead of a single bundle.

matrix/lambdas/daemons/worker.py
```diff
@@ -37,13 +39,28 @@ def run(self, format: str, worker_chunk_spec: dict):
...
-                                      dss_instance=self._deployment_stage)
-        group = zarr.group(store=dss_zarr_store)
-        exp_df, qc_df = convert_dss_zarr_root_to_subset_pandas_dfs(group, self._input_start_row, self._input_end_row)
+        exp_dfs = []
+        qc_dfs = []
+        for chunk_idx in range(len(self._bundle_uuids)):
+            dss_zarr_store = DSSZarrStore(bundle_uuid=self._bundle_uuids[chunk_idx],
+                                          bundle_version=self._bundle_versions[chunk_idx],
+                                          dss_instance=self._deployment_stage)
+            group = zarr.group(store=dss_zarr_store)
```
calvinnhieu: I know you didn't write this code, but I think this should be a responsibility of `convert_dss_zar_root_to_subset_pandas_dfs` and that function should accept a `dss_zarr_store` as opposed to a zarr group (actually the zarr group is available via `dss_zarr_store`)

matrix/lambdas/daemons/driver.py
mckinsel: Yeah, I'll break it out. fwiw, I'm just copying the `grouper` function from [here](https://docs.python.org/3.6/library/itertools.html#itertools-recipes).

matrix/lambdas/daemons/driver.py
mckinsel: Agree. Wasn't sure if you guys had a reason for splitting it up in the driver, but sounds like no, so I'll push it down a bit further.

matrix/lambdas/daemons/mapper.py
mckinsel: :+1:

matrix/lambdas/daemons/mapper.py
mckinsel: Ah right, okay I'll get rid of it.

matrix/lambdas/daemons/mapper.py
mckinsel: Each `worker_chunk_spec` should now be a list of dicts rather than a dict. Is there a way to make that clearer? Maybe call it `worker_chunks_spec`?

matrix/lambdas/daemons/mapper.py
mckinsel: Actually, if it's okay with you, I'd like to push this off to a later PR. Changing it requires some changes to the mocks in the unit tests.

PR #88, opened 2018-10-21
Parthshahva terraform converter

parthshahva:
VPC and Batch service setup for file conversion


terraform/modules/matrix-service/infra/converter_batch.tf
```diff
@@ -0,0 +1,233 @@
...
+  name = "dcp-matrix-converter-job-${var.deployment_stage}"
+  policy = <<POLICY
+{
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Sid": "s3",
+        "Effect": "Allow",
+        "Action": [
+          "s3:*"
```
calvinnhieu: Should restrict

terraform/modules/matrix-service/infra/converter_batch.tf
```diff
@@ -0,0 +1,233 @@
...
+
+resource "aws_batch_job_definition" "converter_job_def" {
+    name = "dcp-matrix-converter-job-definition-${var.deployment_stage}"
+    type = "container"
+    container_properties = <<CONTAINER_PROPERTIES
+{
+  "command": [],
+  "image": "humancellatlas/matrix-converter:1",
+  "memory": 4096,
+  "vcpus": 4,
```
calvinnhieu: Is there a reason we're setting this lower than the min vCPUs on the batch compute environment?

terraform/modules/matrix-service/infra/vpc.tf
```diff
@@ -0,0 +1,137 @@
+//
```
calvinnhieu: Limited experience with VPCs, @mckinsel can I get your review on this?

mckinsel: Looks fine to me. Probably boilerplate terraform for a VPC with public subnets. I'm not sure why we want a subnet for each AZ, but maybe something nice happens with Batch failing over.

PR #89, opened 2018-10-21
matrix conversion code + docker + small flow updates

parthshahva:
Changes to code flow, tests, and docker/batch setup for matrix conversion


requirements.txt
```diff
@@ -10,3 +10,7 @@ zarr==2.2.0
 pandas==0.23.4
```
calvinnhieu: In favor of minimizing deployed package sizes, I've stopped using this top level requirements.txt and specifying requirements specific to each package. Moving to appropriate requirements files

matrix/lambdas/daemons/reducer.py
```diff
@@ -80,11 +84,52 @@ def run(self):
             }
             s3.open(zarray_key, 'wb').write(json.dumps(zarray).encode())
 
+        if self.format != "zarr":
```
calvinnhieu: note to self: should be set on output entry created

PR #90, opened 2018-10-22
Create zarr store abstraction

calvinnhieu:
To provide access properties to zarrays in expression matrices stored in S3

### DIFF
--- a/matrix/common/dss_zarr_store.py
+++ b/matrix/common/dss_zarr_store.py
@@ -8,8 +8,10 @@ import hca
 import zarr
 from hca import HCAConfig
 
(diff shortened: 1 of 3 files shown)


calvinnhieu:
I see! That makes sense - I tried to abstract out the concept of a group thinking it would be a simpler interface, but I see how that oversimplifies the details/is incorrect. Thanks for the explanation - this suggestion looks good! With the demo coming up, I'll table this refactor for later. Will update https://github.com/HumanCellAtlas/matrix-service/pull/91

calvinnhieu:
Closing PR, will reimplement refactor per Marcus' comments

PR #93, opened 2018-10-23
Add pancreas data to functional tests

calvinnhieu:
* Add functional test to test configurable subset of SS2 pancreas data
* Introduce environment variables `MATRIX_TEST_NUM_BUNDLES` and `MATRIX_TEST_TIMEOUT` to configure size of functional test


tests/functional/test_matrix_service.py
```diff
@@ -57,10 +60,25 @@ def test_matrix_service(self):
...
+
+        request_id = self._post_matrix_service_request(bundle_fqids, "zarr")
+
+        # wait for request to complete
+        time.sleep(2)
+        WaitFor(self._poll_get_matrix_service_request, request_id)\
+            .to_return_value(MatrixRequestStatus.COMPLETE.value, timeout_seconds=timeout)
+
+        # TODO: analyze results
+        # self._analyze_matrix_results(request_id)
```
calvinnhieu: Currently this test is just an easy way to configure the size of the pancreas data to test. Any ideas on how we can do some results validation?

mckinsel: Easiest thing would be to check that the number of rows in the output matrix matches `num_bundles`. That would actually catch all the errors we've observed thus far.

PR #94, opened 2018-10-23
parallelize read from DSS in workers

calvinnhieu:


matrix/lambdas/daemons/worker.py
```diff
@@ -41,15 +42,21 @@ def run(self, format: str, worker_chunk_spec: typing.List[dict]):
...
+                                   for chunk_idx in range(len(self._bundle_uuids))}
+            for future in concurrent.futures.as_completed(future_to_chunk_map):
+                chunk_idx = future_to_chunk_map[future]
+                try:
+                    dss_zarr_store = future.result()
+                except Exception as e:
+                    print(f"Reading bundle uuid {self._bundle_uuids[chunk_idx]} from DSS caused exception {e}")
+                    continue
+                group = zarr.group(store=dss_zarr_store)
+                exp_df, qc_df = convert_dss_zarr_root_to_subset_pandas_dfs(
```
mckinsel: I suspect it's going to be important to run `convert_dss_zarr_root_to_subset_pandas_dfs` in parallel. There are a lot of DSS requests that get made in there. Could you move that into the `_read` function?

calvinnhieu: Oh totally missed this. Good catch

matrix/lambdas/daemons/worker.py
```diff
@@ -81,13 +91,22 @@ def run(self, format: str, worker_chunk_spec: typing.List[dict]):
...
             }
             self.lambda_handler.invoke(LambdaName.REDUCER, reducer_payload)
 
+    def _parse_chunk_to_dataframe(self, i: int):
+        dss_zarr_store = DSSZarrStore(bundle_uuid=self._bundle_uuids[i],
+                                      bundle_version=self._bundle_versions[i],
+                                      dss_instance=self._deployment_stage)
+        group = zarr.group(store=dss_zarr_store)
+        if not self.zarr_group:
+            self.zarr_group = group
```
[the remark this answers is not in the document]
mckinsel: Fact check: no it does not.

matrix/lambdas/daemons/mapper.py
```diff
@@ -82,19 +78,13 @@ def _get_chunk_specs(bundle_fqids: typing.List[str]) -> typing.List[dict]:
...
-                                      bundle_version=bundle_version,
-                                      dss_instance=os.environ['DEPLOYMENT_STAGE'])
-
-            root = zarr.group(store=zarr_store)
-
-            rows_per_chunk = root.expression.chunks[0]
-            total_chunks = root.expression.nchunks
+            # TODO: Not this! This is taking advantage of the fact that every
+            # matrixable bundle in the DSS will at first have data for a single
+            # cell. That won't be true for too long
```
calvinnhieu: When do we expect (10x?) bundles to be available in S3?

daemons/driver/Makefile
```diff
@@ -22,7 +22,6 @@ build:
 
 stage: build
 	aws s3 cp $(ZIP_FILE) s3://$(BUCKET)/$(STAGED_FILE_KEY)
-	rm -rf target
```
calvinnhieu: *

matrix/common/dynamo_utils.py
```diff
@@ -115,9 +115,9 @@ def acquire(self):
             # Or maybe we hold the lock ourselves, then just return
             elif db_response["Item"]["LockHolder"] == self._lock_id:
                 return
-
+            print(f"Waiting for lock on {self._lock_key}")
```
calvinnhieu: Does it make sense to use logger here?

matrix/lambdas/daemons/worker.py
```diff
@@ -43,24 +46,31 @@ def run(self, format: str, worker_chunk_spec: typing.List[dict]):
...
+            future_to_chunk_map = {executor.submit(self._parse_chunk_to_dataframe, chunk_idx): chunk_idx
+                                   for chunk_idx in range(num_bundles)}
+            for future in concurrent.futures.as_completed(future_to_chunk_map):
+                chunk_idx = future_to_chunk_map[future]
+                try:
+                    exp_df, qc_df = future.result()
+                except Exception as e:
+                    logger.debug(f"Parsing bundle uuid {self._bundle_uuids[chunk_idx]} from DSS "
+                                 f"to pandas.DataFrame caused exception {e}")
+                    logger.debug(traceback.format_exc())
```
calvinnhieu: Do we want to keep this? I guess it could be helpful, but might flood logs. How often does this Exception get thrown now by the way? It was previously thrown ~30-50% of the time due to shared sessions in the DSSClient

tests/unit/lambdas/daemons/test_mapper.py
```diff
@@ -118,6 +121,8 @@ def test_get_chunk_specs_ok(self, mock_dss_zarr_store, mock_zarr_group):
             self.assertEqual(chunk_spec['start_row'], i * chunk_size)
             self.assertEqual(chunk_spec['num_rows'], chunk_size)
 
+    @pytest.mark.skip(reason="Not needed while one-cell bundle assumption is in place.")
```
calvinnhieu: Can we add a TODO here to address when assumption is no longer true?

matrix/lambdas/daemons/mapper.py
mckinsel: In the DSS? Probably a couple weeks after the demo.

calvinnhieu: Ya *DSS

matrix/common/dynamo_utils.py
mckinsel: Oh yeah, sorry this was something I dropped in and forgot about.

matrix/lambdas/daemons/worker.py
mckinsel: Should never happen now. Every exception should be interesting.

tests/unit/lambdas/daemons/test_mapper.py
mckinsel: Yeah, filed issue #95 as well.

matrix/lambdas/daemons/worker.py
calvinnhieu: Cool cool. Do you know if re raising the exception would be better? I'm only using traceback here to retrieve the stack trace

mckinsel: Oh, yeah probably better to just `raise` after this.

PR #96, opened 2018-10-26
Expose errors during processing to clients via GET endpoint

calvinnhieu:
- Created `RequestTracker` class to abstract request related data and updates
- Added simple error tracking for failed Lambda invocations as example usage
- Updated GET endpoint output errors to client
- Refactored some of `DynamoHandler`
- Wrote tests

### DIFF
--- a/matrix/common/dss_zarr_store.py
+++ b/matrix/common/dss_zarr_store.py
@@ -9,6 +9,8 @@ import hca
 import zarr
 from hca import HCAConfig
 
+from matrix.common.exceptions import MatrixException
+
 
 class DSSZarrStore(MutableMapping):
     """
@@ -88,6 +90,7 @@ class DSSZarrStore(MutableMapping):
                 }
 
         self._root = zarr.group(store=self)
+        self._validate_zarr()
 
     @property
     def bundle_uuid(self):
@@ -139,6 +142,14 @@ class DSSZarrStore(MutableMapping):
 
         client = hca.dss.DSSClient(config=dss_config)
         return client
+
+    def _validate_zarr(self):
+        expected_fields = ['expression', 'cell_id', 'cell_metadata_string', 'cell_metadata_numeric',
+                           'cell_metadata_string_name', 'cell_metadata_numeric_name', 'gene_id', 'gene_metadata',
+                           'gene_metadata_name']
+        if any(field not in self._root for field in expected_fields):
+            raise MatrixException(400, f"Unable to process bundle {self.bundle_uuid}.{self.bundle_version}. "
+                                       f"No expression data found.")
 
     def __setitem__(self, key, value):
         raise NotImplementedError("The HCA Data Storage System is read-only.")
--- a/matrix/common/dynamo_handler.py
+++ b/matrix/common/dynamo_handler.py
@@ -19,6 +19,8 @@ class StateTableField(TableField):
     Field names for State table in DynamoDB.
     """
     REQUEST_ID = "RequestId"
+    EXPECTED_DRIVER_EXECUTIONS = "ExpectedDriverExecutions"
+    COMPLETED_DRIVER_EXECUTIONS = "CompletedDriverExecutions"
     EXPECTED_MAPPER_EXECUTIONS = "ExpectedMapperExecutions"
     COMPLETED_MAPPER_EXECUTIONS = "CompletedMapperExecutions"
     EXPECTED_WORKER_EXECUTIONS = "ExpectedWorkerExecutions"
@@ -36,6 +38,7 @@ class OutputTableField(TableField):
     REQUEST_ID = "RequestId"
     ROW_COUNT = "RowCount"
     FORMAT = "Format"
+    ERROR = "Error"
 
 
 class DynamoTable(Enum):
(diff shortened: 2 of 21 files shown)


calvinnhieu:
I seriously apologize for the large PR haha. If it makes it any better, the commits are quite self contained and it might help to review the changes per commit. Otherwise, I can break this into a multi-PR chain. Let me know which you prefer!

tests/unit/common/test_dynamo_handler.py
```diff
@@ -24,8 +24,8 @@ def setUp(self):
         self.request_id = str(uuid.uuid4())
         self.format = "zarr"
 
-        self.create_test_state_table(self.dynamo)
-        self.create_test_output_table(self.dynamo)
+        self.create_test_state_table()
+        self.create_test_output_table()
```
mckinsel: Can you also remove `self.dynamo` here?

matrix/common/dynamo_handler.py
```diff
@@ -68,27 +71,34 @@ def _get_dynamo_table_resource_from_enum(self, dynamo_table: DynamoTable):
...
+                                 format: str=MatrixFormat.ZARR.value):
         """
         Put a new item in the DynamoDB table responsible for tracking task execution states and
-        counts for a specified job.
+        counts for a specified request.
 
         :param request_id: UUID identifying a filter merge job request.
-        :param num_bundles: Number of bundles to be processed.
+        :param num_mappers: Number of mapper lambdas expected to be invoked.
+        :param num_workers: Number of worker lambdas expected to be invoked.
```
mckinsel: I'm not sure the driver is going to know the number of worker lambdas to be invoked once we have other bundle types.

calvinnhieu: Ah, because we'll chunk out each mapper differently based on its load yea? I'll change it back then, was just trying to clean up

matrix/common/lambda_handler.py
```diff
@@ -26,8 +32,15 @@ def invoke(self, fn_name: LambdaName, payload: dict):
...
-            Payload=json.dumps(payload).encode(),
-        )
+        try:
+            self._client.invoke(
+                FunctionName=fn_name.value,
+                InvocationType="Event",
+                Payload=json.dumps(payload).encode(),
+            )
+        except ClientError:
+            size = len(json.dumps(payload).encore())
```
mckinsel: `encode`?

calvinnhieu: lol clearly didnt test this. thanks good catch

matrix/common/request_tracker.py
```diff
@@ -0,0 +1,108 @@
...
+    @property
+    def format(self) -> str:
+        """
+        The request's user specified output file format of the resultant expression matrix.
+        :return: str The file format (one of MatrixFormat)
+        """
+        if not self._format:
+            self._format = self.dynamo_handler.get_table_item(DynamoTable.OUTPUT_TABLE,
+                                                              self.request_id)[OutputTableField.FORMAT.value]
+        return self._format
```
mckinsel: What value should this have the user doesn't specify a format in their request?

matrix/common/lambda_handler.py
```diff
@@ -26,8 +32,15 @@ def invoke(self, fn_name: LambdaName, payload: dict):
...
+            self.request_tracker.log_error(f"An error occurred while processing your request. The size of the "
```
mckinsel: Probably a bit optimistic to assume that this is always the reason for `ClientError`. If you want a special message, maybe something like
```python
except ClientError as e:
    if e["Error"]["Code"] == "EntityTooLarge":
        # special log
    else:
        log_error(e["Error"]["Message"])
```

matrix/common/request_tracker.py
```diff
@@ -0,0 +1,108 @@
...
+    Maps request subtasks to corresponding state tracking field in DynamoDB.
+    """
+    DRIVER = StateTableField.COMPLETED_DRIVER_EXECUTIONS
+    MAPPER = StateTableField.COMPLETED_MAPPER_EXECUTIONS
+    WORKER = StateTableField.COMPLETED_WORKER_EXECUTIONS
+    REDUCER = StateTableField.COMPLETED_REDUCER_EXECUTIONS
+    CONVERTER = StateTableField.COMPLETED_CONVERTER_EXECUTIONS
+
+
+class RequestTracker:
```
mckinsel: If you want to roll everything into here, I think you'll need a method to increment expected executions. Not needed for handling the one cell SS2 bundles, but maybe file an issue or mark a TODO.

tests/unit/common/test_dynamo_handler.py
calvinnhieu: Oh yea it's removed in a later commit

matrix/common/request_tracker.py
calvinnhieu: Yup, adding now to increment EXPECTED_WORKER... in Mapper

calvinnhieu:
Yea, that's true - this PR isn't capturing the errors that we would want to and agreed that the Lambda invoke issue we're solving anyways. The PR was getting big and it's more just setting it up so we can capture those errors. That's a good start re: invalid bundle fqid. I can add to this PR or open another one after this?

matrix/common/lambda_handler.py
calvinnhieu: Right and as you said, this capturing this error's kind of unnecessary. I couldn't think of another error to capture haha. I'll scrap this and handle the unreadable bundle case then

matrix/common/dss_zarr_store.py
```diff
@@ -137,6 +140,14 @@ def _get_dss_client(self, dss_instance="integration"):
         client = hca.dss.DSSClient(config=dss_config)
         return client
 
+    def _validate_zarr(self):
+        expected_fields = ['expression', 'cell_id', 'cell_metadata_string', 'cell_metadata_numeric',
+                           'cell_metadata_string_name', 'cell_metadata_numeric_name', 'gene_id', 'gene_metadata',
+                           'gene_metadata_name']
```
mckinsel: It seems like this is a repeat of the properties of this class. It would better if that list existed in only one place...though I'm not sure of the best way to do that.

calvinnhieu: Yea, this was one of the use cases for the enum I wanted to introduce with the zarr store/group refactor. I was thinking we'd clean this up when we introduce that refactor

PR #98, opened 2018-10-29
Move url parsing logic into driver

mckinsel:
There's a limit on request size for a lambda, so we can't pass the whole
list of bundle fqids to the driver.

For the mappers and workers, this is less of a concern because we split
up the list of fqids into pretty small chunks.

### DIFF
--- a/matrix/lambdas/api/core.py
+++ b/matrix/lambdas/api/core.py
@@ -1,3 +1,4 @@
+import json
 import os
 import requests
 import uuid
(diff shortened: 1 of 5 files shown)


matrix/lambdas/api/core.py
```diff
@@ -47,24 +46,26 @@ def post_matrix(body: dict):
                                                 "Visit https://matrix.dev.data.humancellatlas.org for more information."
                                  })
 
+    if not has_url and len(body['bundle_fqids']) > 128000:
```
calvinnhieu: Does this check the length of the array or the size in bytes of `body['bundle_fqids]`?

mckinsel: Good call, pretty sure it's the former. I changed it.

PR #118, opened 2018-10-30
Update swagger API documentation

calvinnhieu:
- serve API documentation on route `/`


config/matrix-api.yml
```diff
@@ -4,7 +4,7 @@ info:
   title: HCA Matrix Service
   description: Human Cell Atlas Matrix Service API
   version: 0.0.1
-host: ${API_HOST}
+host: matrix.dev.data.humancellatlas.org
```
parthshahva: why is this hardcoded?

chalice/Makefile
```diff
@@ -30,4 +30,3 @@ deploy: clean build
 	@echo -e "#########################################################\n"
 	./build_deploy_config.sh $(DEPLOYMENT_STAGE)
 	chalice deploy --no-autogen-policy --stage $(DEPLOYMENT_STAGE) --api-gateway-stage $(DEPLOYMENT_STAGE)
-	rm -rf chalicelib
```
mckinsel: What's this bit about?

chalice/app.py
```diff
@@ -49,14 +49,14 @@ def dispatch(*args, **kwargs):
...
-    # with open(os.path.join(pkg_root, "index.html")) as fh:
-    #     swagger_ui_html = fh.read()
-    #
-    # @app.route("/")
-    # def serve_swagger_ui():
-    #     return chalice.Response(status_code=200,
-    #                             headers={"Content-Type": "text/html"},
-    #                             body=swagger_ui_html)
+    with open(os.path.join(pkg_root, "index.html")) as fh:
+        swagger_ui_html = fh.read()
```
mckinsel: Does chalice make the "index.html"?

chalice/Makefile
calvinnhieu: I added this awhile ago to keep the working dir clean, but removing it now to preserve `chalicelib/index.html` (same pattern as other repos)

chalice/app.py
calvinnhieu: No, I looked at Upload's `index.html` and it's a template with a bunch of CSS that populates the content by reading in a Swagger yaml, so I copied it over.

calvinnhieu: I would think that Swagger has a tool to generate the `index.html`, but I don't know where/what that is

PR #119, opened 2018-10-30
Add version and health check endpoints

calvinnhieu:
- `/version`
- `/internal/health`

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -58,6 +58,25 @@ def get_chalice_app(flask_app):
                                 headers={"Content-Type": "text/html"},
                                 body=swagger_ui_html)
 
+    @app.route("/version")
+    def version():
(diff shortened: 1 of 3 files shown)


chalice/app.py
```diff
@@ -58,6 +58,25 @@ def dispatch(*args, **kwargs):
     #                             headers={"Content-Type": "text/html"},
     #                             body=swagger_ui_html)
 
+    @app.route("/version")
+    def version():
```
mckinsel: Isn't version also encoded in the url?

calvinnhieu: Following the pattern of other component version endpoints, this serves the version of code that is deployed, generated by `git describe --tags --always` which includes the hash of the commit and the latest release tag, if available.

PR #120, opened 2018-10-31
Implement caching of requests

mckinsel:
This breaks all the tests.


matrix/common/dynamo_handler.py
```diff
@@ -203,3 +214,36 @@ def _increment_field(self, table, key_dict: dict, field_enum: TableField, increm
             time.sleep(.5)
 
         return start_value, new_value
+
+    def get_request_hash(self, request_id):
```
calvinnhieu: To minimize repeated code, we've been using `get_table_item` to read items from tables. Then, I'd argue `get_request_hash` would live in `RequestTracker`

calvinnhieu: Oh I see now - instead of using `RequestTracker`, I would update `RequestCache.retrieve_hash` to use `get_table_item`

matrix/common/request_cache.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def retrieve_hash(self) -> typing.Union[str, None]:
+        """Look up the hash value for the request.
+
+        Returns None if the request has been initialized but the hash has not yet
+        been set.
+
+        Raises RequestNotFound if the request id isn't in the cache table.
+        """
+        request_hash = self._dynamo_handler.get_request_hash(self._request_id)
```
calvinnhieu: Use `get_table_item` instead?

tests/unit/common/test_request_cache.py
```diff
@@ -0,0 +1,44 @@
...
+            self.request_cache.retrieve_hash()
+
+    def test_set_and_retrieve_hash(self):
+
+        bundle_fqids = ["bundle1.version0", "bundle2.version0"]
+        format_ = "test_format"
+
+        hash_1 = self.request_cache.set_hash(bundle_fqids, format_)
+
+        copy_bundle_fqids = copy.deepcopy(bundle_fqids)
```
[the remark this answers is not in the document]
mckinsel: 👍

mckinsel:
@parthshahva thanks, I built a new image and included the version bump in the PR.

matrix/common/request_cache.py
mckinsel: I think this only works if I parametrize `get_table_item` a lot more. The key name is different for the cache table, and I kind of want different things to happen when ids aren't found.

matrix/common/request_cache.py
```diff
@@ -0,0 +1,88 @@
...
+from matrix.common.logging import Logging
+
+logger = Logging.get_logger(__name__)
+
+# Special value representing no hash. Dynamo doesn't like empty fields.
+NULL_REQUEST_HASH = "null"
+
+
+class RequestIdNotFound(Exception):
+    """Exception for request ids missing the the request cache table."""
```
calvinnhieu: nit: typo

matrix/common/request_cache.py
calvinnhieu: Ohh the key names are different. Gotcha

matrix/lambdas/daemons/driver.py
```diff
@@ -43,21 +43,30 @@ def run(self, bundle_fqids: typing.List[str], bundle_fqids_url: str, format: str
             resolved_bundle_fqids = bundle_fqids
         logger.debug(f"resolved bundles: {resolved_bundle_fqids}")
 
+        request_hash = RequestCache(self.request_id).set_hash(resolved_bundle_fqids, format)
+        logger.debug(f"Calculated and set hash {request_hash} for request {self.request_id}")
+        request_tracker = RequestTracker(request_hash)
+
+        if request_tracker.is_initialized:
+            logger.debug(f"Halting because {request_hash} already exists.")
+            return
```
calvinnhieu: So if the request hash exists, but it failed, we still halt and don't retry the request. Do we want to retry failed requests?

matrix/lambdas/daemons/worker.py
```diff
@@ -79,13 +85,13 @@ def _parse_bundles_to_dataframes(self, bundle_uuids: typing.List[str]):
                     exp_df, qc_df = future.result()
                 except MatrixException as e:
                     self.request_tracker.log_error(e.title)
-                    raise
+                    return None, None
```
calvinnhieu: Since we're not re-raising the exception here, we should log the full exception `e` in this case

tests/functional/test_matrix_service.py
```diff
@@ -42,39 +45,62 @@ def setUp(self):
...
 
     def test_matrix_service_invalid_bundle(self):
         test_bundle_uuids = ["bundle1.version", "bundle2.version"]
-        request_id = self._post_matrix_service_request(bundle_fqids=test_bundle_uuids, format="loom")
-        WaitFor(self._poll_get_matrix_service_request, request_id)\
+        self.request_id = self._post_matrix_service_request(
+            bundle_fqids=test_bundle_uuids, format="loom")
+        WaitFor(self._poll_get_matrix_service_request, self.request_id)\
             .to_return_value(MatrixRequestStatus.FAILED.value, timeout_seconds=60)
+        time.sleep(8)
```
calvinnhieu: Could you explain what these are for?

matrix/lambdas/daemons/driver.py
mckinsel: I'm not sure. I guess we don't know how many failed requests can be retried. I'd _like_ to think it's not too many.

mckinsel: Okay I added a little change that I think should retry. Let me know if that's what you were thinking.

calvinnhieu: Yup that addressed it

PR #121, opened 2018-10-31
Create swagger docs index.html

calvinnhieu:
- serves docs on https://matrix.dev.data.humancellatlas.org/

### DIFF
--- a/chalice/chalicelib/index.html
+++ b/chalice/chalicelib/index.html
@@ -1,0 +1,90 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>DCP Matrix Service API - Swagger Definition</title>
+    <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Source+Code+Pro:300,600|Titillium+Web:400,600,700" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


chalice/chalicelib/index.html
```diff
@@ -0,0 +1,90 @@
...
+</svg>
+
+<div id="swagger-ui"></div>
+
+<script src="https://cdnjs.cloudflare.com/ajax/libs/swagger-ui/3.0.13/swagger-ui-bundle.js" integrity="sha256-aCOXWFczqyopoNlhLgud1jK0sLJ9iLtKyo++l+iZ52Q=" crossorigin="anonymous"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/swagger-ui/3.0.13/swagger-ui-standalone-preset.js" integrity="sha256-PBAt6vT094Le9f8Z2slL0xKSC+xHTfrDSTlwngLe76w=" crossorigin="anonymous"></script>
+<script>
+window.onload = function() {
+  const ui = SwaggerUIBundle({
+    url: window.location.origin + "/v0/swagger.json",
```
mckinsel: I don't know if there's an easy way around this, but I think if we every bump the version in the url, we're likely to forget about the `v0` here.

calvinnhieu: Yea I don't think we can parameterize `v0`. Another option is to serve the file on `/swagger.json` instead, which means removing `basePath` from the swagger spec and explicitly specifying `v0` on each endpoint. I'd vote in favor of keeping `basePath`.

mckinsel: Okay, sounds fine then.

PR #128, opened 2018-11-08
reduce batch resources

parthshahva:
- Reduce job definition for matrix conversion to 2gb RAM and 2cpus
- Reduce min vcpus and desired vcpus to 4


parthshahva:
Should not have impact on conversion time

Issue #39, opened 2018-09-18
Receive csv as a MatrixFormat


brianraymor:
**Use Case: Provide Multiple File Formats**

*When evaluating what file formats people may need, it may be important to provide file formats based on de facto standards given commonly used computational biology tools. In this scenario the user may be able to select what type of files they receive which may not be optimized for performance but are optimized for their workflows.*

**Acceptance Criteria**

- [x] Request "csv" as the matrix format
- [x] zarr is converted to "csv"
- [x] Receive a matrix in the "csv" format
- [x] Functional test to validate conversion

PR #140, opened 2018-11-17
Enable .mtx and .csv formats

calvinnhieu:
Tested on predev environment

.mtx (https://github.com/HumanCellAtlas/matrix-service/issues/39):
Results on predev: https://s3.amazonaws.com/dcp-matrix-service-results-predev/dae8194f299ada5284442893ead8b8f49b4f29a8d8a7f2e9c513e9971378c628.mtx.gz

.csv (https://github.com/HumanCellAtlas/matrix-service/issues/59):
Results on predev: https://s3.amazonaws.com/dcp-matrix-service-results-predev/7da730847f594d33c136c84d4d0701adedbe6933f2c1783435452595a5be9502.csv.gz


calvinnhieu:
@mckinsel Could you verify the output results?

mckinsel:
Hey the output from `mtx` is actually a gzipped tarball. Should we change the format name to reflect that?

calvinnhieu:
Not sure. For consistency with the other format names I want to keep it the same, but for clarity/UX it makes sense to change the input to something like `mtx.gz`. Do you think this would cause confusion if we don't change it? Note that the `matrix_location` URL is updated to return `mtx.gz`

mckinsel:
I think a user would end up doing something like `wget https://somethingsomething.mtx` and then have  a local file with an mtx extension that they need to unzip and untar. That would probably be a surprise.

brianraymor:
It's surprising that there's not a standard for tar.gz. Pick your adventure: https://superuser.com/questions/901962/what-is-the-correct-mime-type-for-a-tar-gz-file 

In the wget scenario, even if we advertised *mtx+gzip*, users on Windows are going to be surprised by the .tar format unless they've installed cygwin?

mckinsel:
Maybe we should just zip it, and then call it `mtx.zip`?

brianraymor:
I like *mtx+gzip* a'la - https://tools.ietf.org/html/rfc8460#section-6.3 but I appreciate it's a [bike shed](http://bikeshed.com/).

calvinnhieu:
@mckinsel That sounds good - it's simple, works on Windows and achieves the same thing as tar and gz. Does it make sense to also use zip for csv? Also if we're compressing loom files, we should update the return file extension to reflect that (I'm not sure if we are)

mckinsel:
So zip is what the Single Cell Expression Atlas does, so I'd lean toward that. @brianraymor I'm not sure where we'd set the content-type. Do you mean when we put the object in s3?

brianraymor:
What Content-Type(s) are returned when a client GETs the *matrix_location* URL?

calvinnhieu:
@brianraymor Looks like `binary/octet-stream`
screen shot 2018-11-19 at 4 42 15 pm

mckinsel:
Pretty sure that's just the default. We can set it when we PUT.

PR #143, opened 2018-11-21
Create batch handler

calvinnhieu:
- refactor Reducer, create BatchHandler
- added tests

### DIFF
--- a/tests/unit/__init__.py
+++ b/tests/unit/__init__.py
@@ -5,7 +5,6 @@ import boto3
 
 from moto import mock_dynamodb2, mock_s3
 
-os.environ['ACCOUNT_ID'] = "test_account_id"
 os.environ['DEPLOYMENT_STAGE'] = "dev"
 os.environ['AWS_DEFAULT_REGION'] = "us-east-1"
 os.environ['AWS_ACCESS_KEY_ID'] = "ak"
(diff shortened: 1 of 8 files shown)


tests/unit/__init__.py
```diff
@@ -16,6 +15,8 @@
 os.environ['DYNAMO_CACHE_TABLE_NAME'] = f"dcp-matrix-service-cache-table-{os.environ['DEPLOYMENT_STAGE']}"
 os.environ['S3_RESULTS_BUCKET'] = f"dcp-matrix-service-results-{os.environ['DEPLOYMENT_STAGE']}"
 os.environ['DYNAMO_LOCK_TABLE_NAME'] = f"dcp-matrix-service-lock-table-{os.environ['DEPLOYMENT_STAGE']}"
+os.environ['BATCH_CONVERTER_JOB_QUEUE_ARN'] = f"test-job-queue"
+os.environ['BATCH_CONVERTER_JOB_DEFINITION_ARN'] = f"test-job-definition"
```
calvinnhieu: I was considering the removing dependency on environment variables here (by requiring `source environment` to be run before running unit tests), however I think it makes sense that unit tests are decoupled from the deployed state and that this is fine. Following this, I would opt to update these to "dummy" values (in another PR).

calvinnhieu:
Functional tests passing with full deploy to predev

PR #145, opened 2018-11-26
Trivialize unit test environment variables

calvinnhieu:
I believe unit test environment variables should not be associated with their deployment values. Organizationally, this makes it clearer that these values don't need to be updated when the real values change, however, they still need to be set.


calvinnhieu:
Force push to rebase master ^

PR #151, opened 2018-11-30
Increase unit test coverage to 90%

calvinnhieu:
- ~A few more tests to write for `s3_zarr_store.py` that should bring us to 90%~
- Commits are self-contained and can be reviewed independently if that makes it more manageable


matrix/common/aws/dynamo_lock.py
```diff
@@ -1,23 +1,22 @@
-"""TODO TEST AND/OR REFACTOR THIS CLASS OUT OF IMPLEMENTATION.
+"""TODO: REFACTOR THIS CLASS OUT OF IMPLEMENTATION.
```
parthshahva: We should just remove this todo out of the code. We can create a spike ticket to investigate options that don't involve dynamo lock

matrix/docker/matrix_converter.py
```diff
@@ -13,8 +13,11 @@
 import scipy.io
```
parthshahva: Since this PR makes changes to file conversion, it needs to be accompanied with a rebuild of the docker image at https://github.com/HumanCellAtlas/matrix-service/blob/master/docker-images/matrix-converter/Makefile#L20 and increment of image version in use for job definition at https://github.com/HumanCellAtlas/matrix-service/blob/master/terraform/modules/matrix-service/infra/converter_batch.tf#L216

matrix/common/aws/dynamo_lock.py
brianraymor: Yes please. Let's please avoid **TODO** in code and create tracking issues as necessary for visibility.

calvinnhieu: Cool, created https://app.zenhub.com/workspaces/dcp-backlogs-5ac7bcf9465cb172b77760d9/issues/humancellatlas/matrix-service/159

PR #161, opened 2018-12-06
Add ignore_cache option to POST /matrix

calvinnhieu:
**Changes**
These changes add the input parameter `ignore_cache` to the POST /matrix endpoint. The parameter determines the following behavior in the Driver:
- if `True`, `RequestHash._hash_request` will non-deterministically compute a request hash to avoid a cache hit
- if `False`, `RequestHash._hash_request` will compute the request hash normally (deterministically)

**Design**
Currently, there is a need to test the service independent of its caching mechanism in order to test all of its components and gather comprehensive performance metrics. Furthermore, not all errors are captured during processing which may result in a "hanging" request stuck in the "In Progress" state. The `ignore_cache` flag solves these issues by providing a bypass to the caching mechanism to:

1) Run tests independent of the caching mechanism
2) Rerun "stuck" or corrupt requests

An alternative solution to resolve (1) could be to delete requests issued by tests which is the current implementation. This is a sufficient solution, however, the added benefit of (2) I believe justifies the value of the flag.

### DIFF
--- a/config/matrix-api.yml
+++ b/config/matrix-api.yml
@@ -25,6 +25,8 @@ definitions:
         format: url
       format:
         $ref: '#/definitions/MatrixFormat'
+      ignore_cache:
+        type: boolean
   MatrixFormat:
     type: string
     enum:
--- a/daemons/driver/app.py
+++ b/daemons/driver/app.py
@@ -4,7 +4,7 @@ from matrix.lambdas.daemons.driver import Driver
 def driver_handler(event, context):
     # TODO: better error handling
     assert ('request_id' in event and 'format' in event and 'bundle_fqids' in event and
-            'bundle_fqids_url' in event)
+            'bundle_fqids_url' in event and 'ignore_cache')
     assert bool(event["bundle_fqids"]) != bool(event["bundle_fqids_url"])  # xor these
     driver = Driver(event['request_id'])
-    driver.run(event['bundle_fqids'], event["bundle_fqids_url"], event['format'])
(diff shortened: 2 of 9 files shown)


brianraymor:
Why would we expose and document a public API for an internal use case? In addition, there are [cache-control](https://tools.ietf.org/html/rfc7234#section-5.2) headers with no-cache options.

calvinnhieu:
Is this exclusively an internal use case? I think it's reasonable for clients requesting a matrix to `ignore_cache`. What do you guys think?

brianraymor:
As described, it's an internal use case. The service should not be serving stale resources which is one rationale for no-cache directives. 

I could also imagine an attack on the service by a client which made the same request over and over again with `ignore_cache` set. For example - https://securityintelligence.com/dissecting-hacktivists-ddos-tool-saphyra-revealed/

calvinnhieu:
I see. Would you recommend keeping the no-cache directive and hiding it behind a non-public API (potentially requiring authentication)? Or a different solution that does not involve the API (e.g. an internal tool to clear the cache)?

mckinsel:
I'm not so sure this addresses use case (2). The request would remain forever stuck unless you always used `ignore_cache`.

calvinnhieu:
I agree it's not a long-term solution for addressing unhandled errors/failures. For (2), this would be a one-time workaround. Maybe a non-public endpoint to clear the cache of a specific request hash (or set of input parameters) would be more appropriate. This could be used in tests to resolve (1), and we plan to resolve (2) using SQS to build more comprehensive error/failure handling

calvinnhieu:
Decision to revise solution per discussion in this thread

Issue #105, opened 2018-10-29
Create Grafana dashboard for CloudWatch metrics


calvinnhieu:
TODO: define which metrics should be captured

parthshahva:
- Read/write on dynamo tables
- API response code breakdown
- Batch latency
- Size of batch queues
- Invocation and Failure count for each daemon lambda
- Invocation and Failure count for api lambda
- Average and max duration for each daemon lambda
- Failed requests (use dynamo state table to figure this out)
- Number of cache hits vs misses (Will figure out how to measure this after caching goes in)
- Average time from initial post request to completed job (Not sure how to measure this yet)

@mckinsel @calvinnhieu comment here if there is anything else you think of to measure

parthshahva:
MVP grafana dashboard for dev: https://metrics.dev.data.humancellatlas.org/d/matrix-dev/matrix-dev?refresh=1m&panelId=18&orgId=1&from=1543851434706&to=1543860895863

Left: publish custom cloudwatch metrics to measure job and batch stats (not available currently from cloudwatch metrics)

parthshahva:
https://github.com/HumanCellAtlas/dcp-monitoring/pull/19 open for review and https://github.com/HumanCellAtlas/matrix-service/pull/163 is merged. Once #163 is promoted up through prod (integration currently in progress) and https://github.com/HumanCellAtlas/dcp-monitoring/pull/19 is merged in, this ticket can be marked as done. Estimate: EOD 11/10

PR #163, opened 2018-12-07
v1 of custom cloudwatch metrics from matrix service

parthshahva:
This solves the second part of https://github.com/HumanCellAtlas/matrix-service/issues/105. After this is merged in, we will pull these custom cloudwatch metrics into grafana to close out ticket 105.

This PR introduces CloudwatchHandler with function put_metric_data to establish custom cloudwatch metric datapoints for the matrix service


matrix/lambdas/api/core.py
```diff
@@ -59,14 +63,16 @@ def post_matrix(body: dict):
 
     request_id = str(uuid.uuid4())
     RequestCache(request_id).initialize()
-
+    cloudwatch_handler.put_metric_data(
+        metric_name=MetricName.REQUEST,
+        metric_value=1
+    )
```
calvinnhieu: Thoughts on coupling these with specific functions where appropriate? e.g. putting this one in `RequestCache.initialize`. Though it's true this doesn't make sense for all calls to `put_metric_data`, I think the coupling is cleaner and easier to maintain.

matrix/lambdas/daemons/reducer.py
```diff
@@ -26,5 +28,14 @@ def run(self):
 
         if self.request_tracker.format != MatrixFormat.ZARR.value:
             self.batch_handler.schedule_matrix_conversion(self.request_tracker.format)
+            self.cloudwatch_handler.put_metric_data(
+                metric_name=MetricName.CONVERSION_REQUEST,
+                metric_value=1
+            )
```
calvinnhieu: Example of other comment

matrix/common/aws/cloudwatch_handler.py
```diff
@@ -0,0 +1,40 @@
...
+import boto3
+
+
+class MetricName(Enum):
+    """
+    MetricNames for cloudwatch custom metric datapoints.
+    """
+    REQUEST = "Matrix Request"
+    REQUEST_COMPLETION = "Matrix Request Completion"
+    REQUEST_ERROR = "Matrix Request Error"
```
calvinnhieu: Though we do plan on replacing `RequestTracker.log_error` in the future, that could be a good place to start counting error metrics

Issue #131, opened 2018-11-12
DSS/Matrix interaction improvement


parthshahva:
Currently, the interaction between DSS and the Matrix Service is the most fragile part of the system. We are not well protected against dss issues or latency.  Zarr stores for 1 billion cells will only take up 4 tb according to @mckinsel.  We have some options here:

1) Register a subscription against dss and maintain our own copy of zarr stores. Issues here on how to handle updates and deletes

2) Dig directly into s3 rather than going via dss since we share the same aws account (icky)

Action item: Schedule 1 hour design session to discuss prototypes we can build

brianraymor:
Not required for the refactored architecture.

PR #165, opened 2018-12-10
Add short design doc discussing handling of 10x bundles

### LINKED ISSUE #95: Revise bundle and chunk distribution in mapper
In order to improve performance, we took advantage of the fact that each SS2 bundle has only one cell. So, the work that the mapper currently does is pretty trivial: it just distributes a fixed set of bundle uuids to each worker. But 10x bundles will violate that assumption, and the mapper is going to have to check for bundle size and distribute some number of chunks to each worker.

mckinsel:
This fixes #95 , a spike ticket.

### DIFF
--- a/docs/10x_mapper_design.md
+++ b/docs/10x_mapper_design.md
@@ -1,0 +1,83 @@
+# Adjusting mappers and workers to handle 10x bundles
+
+Recall that in the current driver/mapper/worker/reducer scheme, we have a few
+design properties that we want to optimize:
+
+1. *Minimize number of lambda executions.* Each lambda has small chance of
+   failing non-deterministically, and we don't handle that very well. Also each
+   lambda has some overhead.
+2. *Minimize number of parallel writers.* When many workers are writing in
+   parallel to a zarr store, there's a lot of lock contention that can
+   eventually overwhelm the lock table
+3. *Minimize DSS calls.* API calls to the DSS are slow and the distribution of
+   response times has a very long, fat tail. We also have to carefully wrap
+   every interaction with retry logic.
+4. *Never overload a lambda.* If a lambda OOMs or times out, presently our
+   whole query breaks. Moreover, even if a lambda completes in, say, 14
+   minutes, that's a long time for a user to be waiting.
+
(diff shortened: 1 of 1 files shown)


calvinnhieu:
Favor option (2) over (1) given the DSS latency we're already experiencing in the Worker. +1 to subscribing to DSS notifications.

I'm curious what complexity would be introduced if we were to create a job that scrapes all matrix-able bundles from a DSS environment and initializes `bundle_cell_count_table`. On the other hand, such a tool would only improve processing times for the first request and probably isn't worth all the overhead. Option (2) proposed as is (with DSS subscriptions) sounds simple and good to me!

calvinnhieu:
A larger discussion, but relevant here: Do we want to reduce DSS interaction in the Worker as well? We have discussed in the past storing more bundle data than just `bundle_cell_count` outside of the DSS (e.g. all expression matrix data). This would be a complex feature requiring significant engineering effort, however, breaking up the DSS API and distributing its traffic may be necessary for improving performance across the DCP. That being said, I'm unsure how much of our performance bottleneck is due to DSS latency. There are 2 things I'd be interested in diving deeper into from here:
1. What does our performance look like now and where do we need to improve?
    a. This will be captured in the Locust user load tests 
2. Is there a larger, DCP-wide need to reduce stress on the DSS API?
    a. What improvements can we make to the DSS to achieve this? As mentioned above, an example would be a separate API/system to host/serve expression matrix data

parthshahva:
On further thought, option 2 by itself also gives me pause. A dss notification is not something easy to handle operationally. For one, you would also need to handle bundle deletes and updates. There is also the risk (guarantee at large enough scale) that a notification falls through the cracks. You are now also having to maintain another db of sorts. In this case, would we also then code in option 1 if we can't find the bundle in our local table? I think this discussion has a certain level of complexity that warrants an in-person design session. We should combine this discussion with the spike in https://github.com/HumanCellAtlas/matrix-service/issues/131 as there is quite a bit of overlap.

parthshahva:
@mckinsel can this PR be closed per the discussion that followed?

PR #171, opened 2018-12-15
Add user load testing design doc

calvinnhieu:
Created this doc to capture design decisions made pertaining to _Capturing metrics_ and _Visualizing results_. These decisions are a pivot from the previous approach where the tests were consuming results produced by Locust and generating `matplotlib` plots. There are a couple of issues with that approach:
- Locust results are insufficient as they only provide aggregations (average, maximum and minimum) as opposed to individual data points
- the current `matplotlib` plots are simplistic, inflexible and unaligned with our current metrics solutions (CloudWatch, Grafana)

Looking for your input on the proposal and suggestions for other ways to monitor the overall performance of the system!

### DIFF
--- a/docs/user_load_testing.md
+++ b/docs/user_load_testing.md
@@ -1,0 +1,86 @@
+# User load testing
+
+In order to prepare for Closed Beta and General Availability, the Matrix Service is in need of a solution to profile its
+performance in terms of real world usage. A user load testing framework that can measure performance across a number of
+the service's control inputs can achieve this. The following provides concrete definitions for the terms _performance_ 
+and _control inputs_:
+
+### Performance
+
+Performance, in context of the User Load Tests, refers to the processing time of a matrix request measured from the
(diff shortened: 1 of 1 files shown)


docs/user_load_testing.md
```diff
@@ -0,0 +1,84 @@
+# User load testing
+
+In order to prepare for Closed Beta and General Availability, the Matrix Service is in need of a solution to profile its
+performance in terms of real world usage. A user load testing framework that can measure performance across a number of
+the service's control inputs can achieve this. The following provides concrete definitions for the terms _performance_ 
+and _control inputs_:
+
+### Performance
+
+Performance, in context of the User Load Tests, refers to the processing time of a matrix request measured from the
```
parthshahva: Refer to caching here and relevance to the concurrent user test?

calvinnhieu: Sure thing (currently being handled by selecting a random subset of matrixable bundles of the requested input size to lower the chance of a cache hit)

PR #172, opened 2018-12-17
Fix csv output

### LINKED ISSUE #167: Issues with csv download
When unzipped:

1. What is the rationale for two levels of subdirectories - `./tmp/tmpx5k4bgls/matrix.csv.zip`?
2. `matrix.csv.zip` should be named `matrix.csv`
3. This should be consistent with the .mtx directory structure - `./matrix`

Why was this not detected during testing?

mckinsel:
Previously, csv outputs were being created using the `to_csv` from
pandas, which gives you less control over what the compressed file
looks like. In particular, the name of the matrix csv file in the
archive was `/tmp/something/matrix.csv`, which means people would
extract it and never find the matrix. This change sets the archive name
directly.

Fixes #167 .


matrix/docker/matrix_converter.py
```diff
@@ -87,12 +87,16 @@ def zarr_to_csv(zarr_root):
...
+    csv_path = os.path.join(temp_dir, "matrix.csv")
+    csv_zip_path = os.path.join(temp_dir, "matrix.csv.zip")
 
     dataframe = pandas.DataFrame(zarr_root.expression[:], index=zarr_root.cell_id[:],
                                  columns=zarr_root.gene_id[:])
-    dataframe.to_csv(csv_path, compression="zip")
-    return csv_path
+    dataframe.to_csv(csv_path)
+    with zipfile.ZipFile(csv_zip_path, mode="w", compression=zipfile.ZIP_DEFLATED) as z:
+        z.write(csv_path, arcname=os.path.basename(csv_path))
```
calvinnhieu: Just checking, this writes out a `.csv.zip` file?

mckinsel: Yeah, it zips the file at `csv_path`.

mckinsel:
@parthshahva , I bumped the value in `VERSION`, thanks.

Issue #125, opened 2018-11-05
Implement big red button 


parthshahva:
Requirement: Easy way to stop any requests from being made to the matrix service.

Unresolved: Should we cancel all currently running lambda/batch jobs?

brianraymor:
Script that makes the bucket private and disables all the lambdas. Undeploy the API stage.  Validate with Tony.

PR #173, opened 2018-12-17
script to take down matrix service api, lambdas, and job queue

parthshahva:
Big red button script for https://github.com/HumanCellAtlas/matrix-service/issues/125

### DIFF
--- a/scripts/take_down_deployment_be_careful.py
+++ b/scripts/take_down_deployment_be_careful.py
@@ -1,0 +1,49 @@
+import argparse
+
+import boto3
+
+lambda_client = boto3.client('lambda')
+api_gateway_client = boto3.client('apigateway')
+batch_client = boto3.client('batch')
+
+cdn_non_prod_template = "matrix.$DEPLOYMENT_STAGE.data.humancellatlas.org"
+cdn_prod = "matrix.data.humancellatlas.org"
+lambda_name_templates = [
+    "matrix-service-api-$DEPLOYMENT_STAGE",
+    "dcp-matrix-service-driver-$DEPLOYMENT_STAGE",
+    "dcp-matrix-service-mapper-$DEPLOYMENT_STAGE",
+    "dcp-matrix-service-worker-$DEPLOYMENT_STAGE",
+    "dcp-matrix-service-reducer-$DEPLOYMENT_STAGE"
+]
+job_queue_template = "dcp-matrix-converter-queue-$DEPLOYMENT_STAGE"
+
+
+def main(args):
+    # Take down API Gateway
+    if args.environment == "prod":
+        cdn = cdn_prod
+    else:
+        cdn = cdn_non_prod_template.replace("$DEPLOYMENT_STAGE", args.environment)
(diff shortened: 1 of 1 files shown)


scripts/take_down_deployment_be_careful.py
```diff
@@ -0,0 +1,49 @@
...
+        cdn = cdn_non_prod_template.replace("$DEPLOYMENT_STAGE", args.environment)
+    api_gateway_client.delete_domain_name(
+        domainName=cdn
+    )
+    print(f"took down cdn {cdn}")
+
+    # Throttle all involved lambdas
+    for lambda_template in lambda_name_templates:
+        lambda_name = lambda_template.replace("$DEPLOYMENT_STAGE", args.environment)
+        lambda_client.put_function_concurrency(FunctionName=lambda_name, ReservedConcurrentExecutions=0)
```
mckinsel: The AWS docs say "Note that Lambda automatically reserves a buffer of 100 concurrent executions for functions without any reserved concurrency limit.". Does setting this to zero just mean that the functions can use the account-wide limit instead?

mckinsel: :+1:

parthshahva: After invoking this function with value of 0, if you go the UI, you will see the message "Function is throttled" at the top

scripts/take_down_deployment_be_careful.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def main(args):
+    # Take down API Gateway
+    if args.environment == "prod":
+        cdn = cdn_prod
+    else:
+        cdn = cdn_non_prod_template.replace("$DEPLOYMENT_STAGE", args.environment)
+    api_gateway_client.delete_domain_name(
+        domainName=cdn
```
mckinsel: Is it possible for someone to continue to access the API without the custom domain name?

mckinsel:
Looks good. One question: a big part of the current API is just downloading S3 URLs. Those all stay intact after this script is run. Do we want to add a little section that revokes `s3:GetObject` on the results bucket?

scripts/take_down_deployment_be_careful.py
parthshahva: I don't think this is possible. Custom domain name apigateway record is the link between an api gateway stage and route 53. As far as I know, you can't access an api gateway stage without this routing setup

parthshahva: ^I may be wrong here. I didn't know about api gateway invoke urls... Looking into this

calvinnhieu: As @mckinsel brought up, the API is still available via the generated invoke URL `<stage_id>.execute-api.us-east-1.amazonaws.com`  We could attach a [Resource Policy](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-control-access-policy-language-overview.html) to the API denying `invoke` if we want to shut this down as well.

PR #175, opened 2018-12-21
Capture request duration in Cloudwatch Metrics

calvinnhieu:
This work implements _Capturing metrics_ defined in this [doc](https://github.com/HumanCellAtlas/matrix-service/blob/0da63baea58d6f119bc92e29fd28702690c6ea07/docs/user_load_testing.md)

Changes:
- request `CreationDate` is stored in DynamoDB Cache table during POST request
- `request_id` is propagated to all subtasks in order to provide access to `CreationDate` in Cache table
- `Duration` Cloudwatch Metric is published at the ends of the Reducer and Converter subtasks 
    - metric includes `Number of Bundles` and `Output Format` dimensions

Do we also want to track duration metrics for cache hits? This is not included in these changes as I think it would dilute the results that we actually care about (full request processing times).


calvinnhieu:
Most of the changed files are from propagating `request_id` through

matrix/common/aws/dynamo_handler.py
```diff
@@ -148,18 +153,23 @@ def write_request_error(self, request_hash: str, message: str):
             ExpressionAttributeValues={':m': message}
         )
 
-    def get_table_item(self, table: DynamoTable, request_hash: str):
+    def get_table_item(self, table: DynamoTable, request_hash: str="", request_id: str=""):
```
parthshahva: This looks a little messy to be combining request_hash and request_id params alongside the check in line 165. Why not have two separate functions that makes the logic clearer?

matrix/common/request/request_cache.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, request_id: str) -> None:
         self._dynamo_handler = DynamoHandler()
         self._cloudwatch_handler = CloudwatchHandler()
 
+        self._creation_date = ""
```
parthshahva: Why do we need to instantiate an empty string?

matrix/common/request/request_tracker.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, request_hash: str):
         Logging.set_correlation_id(logger, request_hash)
 
         self.request_hash = request_hash
+        self._num_bundles = None
```
parthshahva: Why instantiate this instead of just using the property?

calvinnhieu: This variable caches the value in the table

matrix/common/aws/dynamo_handler.py
calvinnhieu: I agree having two separate primary keys is messier, but I chose to build on top of the same function since we're already selecting the `DynamoTable`. I think having `get_table_item` and `get_cache_table_item` would also be messy. Alternatively, I also considered making the choice between `request_hash` and `request_id` implicit (determined by `DynamoTable`) and the parameter could be called `primary_key`. Thoughts/preferences?

matrix/common/request/request_cache.py
calvinnhieu: Variable isn't recognized as a member unless defined in `__init__`

matrix/common/aws/dynamo_handler.py
parthshahva: This makes sense. I'll leave this up to your judgement

calvinnhieu: Added error handling to current implementation to ensure correct usage

PR #178, opened 2019-01-11
Transpose .mtx formats to store cells by column

calvinnhieu:
https://app.zenhub.com/workspaces/dcp-backlogs-https://github.com/HumanCellAtlas/matrix-service/issues/170


matrix/docker/matrix_converter.py
```diff
@@ -122,7 +122,7 @@ def zarr_to_mtx(zarr_root):
     mtx_path = os.path.join(temp_dir, "matrix.mtx")
     dataframe = pandas.DataFrame(zarr_root.expression[:])
 
-    sparse_mat = scipy.sparse.coo_matrix(dataframe.values)
+    sparse_mat = scipy.sparse.coo_matrix(dataframe.values).transpose()
```
parthshahva: Rather than just calling .transpose() in both the code and tests, it would be nice if we could formulate some type of assertion in the tests for this code change

PR #188, opened 2019-02-11
remove glioblastoma bundles from tests

calvinnhieu:
Functional tests passing (dev, integration)


mckinsel:
Looks good, what did you replace them with?

calvinnhieu:
Latest test SS2 bundles in integration. I believe they're whitelisted @parthshahva

mckinsel:
Is there anything in a cache or results bucket we need to delete?

calvinnhieu:
Ah, yeah. We should clear the predev, dev and integration caches and results buckets - I'll do that now

mckinsel:
Thank you!

calvinnhieu:
Cleared

PR #196, opened 2019-02-14
create redshift resource in dev

calvinnhieu:


terraform/modules/matrix-service/infra/redshift.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_redshift_cluster" "default" {
+  cluster_identifier = "dcp-matrix-service-cluster-${var.deployment_stage}"
+  database_name      = "${var.deployment_stage}"
```
parthshahva: please namespace the database_name with matrix-service as well

terraform/envs/dev/terraform.tfvars.example
```diff
@@ -3,3 +3,5 @@ account_id = ""
 aws_region = ""
 converter_cluster_ami_id = ""
 vpc_cidr_block = ""
+redshift_username = ""
```
parthshahva: Where are the tfvars stored now? Is this change made in this PR?

calvinnhieu: In the S3 terraform bucket, together with the tfstate files

Changes:
https://github.com/HumanCellAtlas/matrix-service/commit/0fc0cb4890f99874f86b55916a8dbc8719ccd874

terraform/modules/matrix-service/infra/redshift.tf
calvinnhieu: Will be named matrix_service_*, does not allow '-'

parthshahva: sounds good

terraform/envs/dev/terraform.tfvars.example
parthshahva: got it. This pattern looks familiar ;)

PR #200, opened 2019-02-22
Implement Redshift initialization data loader

calvinnhieu:
Components:
- `init_cluster.sh`: provisions and executes `init_cluster.py` on an EC2
- `init_cluster.py`: implements `dcplib.etl` to ETL DSS data into Redshift
- `transformers/*`: transforms bundle metadata/analysis data into related Redshift table rows
    - see `MetadataToPsvTransformer` for interface definition
    - Transformers are grouped by data dependencies across tables

TODO:
- unit tests


scripts/redshift/init_cluster.py
```diff
@@ -0,0 +1,223 @@
...
+    Creates tables and loads PSVs in S3 into Redshift via SQL COPY.
+    """
+    conn_url = json.loads(boto3.client('secretsmanager', region_name='us-east-1').
+                          get_secret_value(SecretId='dcp/matrix/dev/database')['SecretString'])['database_uri']
+    url_parts = urlparse(conn_url)
+    dbname = f"matrix_service_{DEPLOYMENT_STAGE}"
+
+    conn_string = f"dbname={dbname} user={url_parts.username} password={url_parts.password} " \
+                  f"host={url_parts.hostname} port={url_parts.port}"
+    conn = psycopg2.connect(conn_string)
```
calvinnhieu: TODO: just use secret string when string points to dev db

scripts/redshift/transformers/__init__.py
```diff
@@ -0,0 +1,63 @@
+import os
+import typing
+from enum import Enum
+
+
+class TableName(Enum):
```
mckinsel: Do you think there's a way to have a single, authoritative record of the tables somewhere? Just trying to minimize the number of things we need to remember to change if we update the schema.

calvinnhieu: I'm on board with that. This record would have two requirements: 1) programmatically readable and 2) accessible from *. In code as an `Enum` satisfies (1), and as far as I can tell, * is the loader/notifications script. Do you think a globally accessible Enum in the codebase works? Or were you thinking the source of truth would be out of the code somewhere?

PR #201, opened 2019-02-26
terraform changes

parthshahva:
terraform changes to bring master up to speed


terraform/modules/matrix-service/infra/s3.tf
```diff
@@ -47,3 +47,18 @@ output "deployment_bucket_id" {
 output "results_bucket_arn" {
   value = "${aws_s3_bucket.matrix-results.arn}"
 }
+
+resource "aws_s3_bucket" "matrix_queries" {
```
calvinnhieu: For passing queries that exceed SQS message size limits?

parthshahva: yup

Issue #181, opened 2019-01-17
NoSuchKey error when requesting csv download


brianraymor:
https://prod.data.humancellatlas.org/explore/projects?filter=%5B%7B%22facetName%22%3A%22project%22%2C%22terms%22%3A%5B%22Single+cell+transcriptome+analysis+of+human+pancreas%22%5D%7D%2C%7B%22facetName%22%3A%22fileFormat%22%2C%22terms%22%3A%5B%22matrix%22%5D%7D%5D

returns:

~~~
<Error>
<Code>NoSuchKey</Code>
<Message>The specified key does not exist.</Message>
<Key>
fc0286efce8dc50d51fdf33c4c6eaee16a69def0a4ca89f4b41755d8a15cee56.csv.zip
</Key>
<RequestId>34F90DCCEB8B6DE3</RequestId>
<HostId>
chNBP5flxU+I8D+mhPGYPukL0dK2n1hPf5Er7pyGPEzYQ3Q+NT+JZuJMlY2STn4MWJOrpPtlmps=
</HostId>
</Error>
~~~

brianraymor:
Don't expect this to reproduce in the new implementation (which does not cache).

brianraymor:
Re-opening because this impacted a Beta user.

PR #206, opened 2019-02-28
Fix broken cache, implement timeouts

calvinnhieu:
**Cache Invalidation**
With these changes, a request in the cache can be invalidated by writing an Error value to the Output table. Doing so will result in:
- Failed returned on GET(request_id) and [fix NoSuchKey error](https://github.com/HumanCellAtlas/matrix-service/issues/181)
- Request can be re-run by resubmitting the POST request (currently forever stuck In Progress)

**Other cache fixes:**
- removed expiration bucket policy on S3 results bucket (source of NoSuchKey error)
- requests older than 1 hour will return Failed on GET instead of In Progress

### DIFF
--- a/matrix/common/request/request_cache.py
+++ b/matrix/common/request/request_cache.py
@@ -1,9 +1,12 @@
 import hashlib
 import typing
+from datetime import timedelta
 
+from matrix.common import date
 from matrix.common.aws.dynamo_handler import DynamoHandler, DynamoTable, CacheTableField
 from matrix.common.aws.cloudwatch_handler import CloudwatchHandler, MetricName
 from matrix.common.logging import Logging
+from matrix.common.request.request_tracker import RequestTracker
 
 logger = Logging.get_logger(__name__)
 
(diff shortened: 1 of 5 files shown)


matrix/common/request/request_cache.py
```diff
@@ -64,6 +67,14 @@ def creation_date(self):
             self._creation_date = item[CacheTableField.CREATION_DATE.value]
         return self._creation_date
 
+    @property
+    def timeout(self) -> bool:
+        timeout = date.to_datetime(self.creation_date) > (date.get_datetime_now()) - timedelta(hours=1)
```
mckinsel: I always get this wrong, but are you certain this is `>`?

matrix/common/request/request_cache.py
```diff
@@ -64,6 +67,14 @@ def creation_date(self):
...
+        if timeout:
+            RequestTracker(self.retrieve_hash()).log_error("This request has timed out after 1 hour."
+                                                           "Please try again by resubmitting the POST request.")
```
mckinsel: Would you need to update the driver for this to work?

matrix/common/request/request_cache.py
calvinnhieu: Nope, good call it's the other way. Adding a test

matrix/common/request/request_cache.py
calvinnhieu: Nope, the driver as is will re-run the request if the RequestHash has an error

PR #199, opened 2019-02-19
initial setup of query runner docker image and ecs/fargate terraform setup

parthshahva:

Issue #194, opened 2019-02-13
Worker infrastructure to run long queries and output results to s3


brianraymor:
When a request for a matrix is received, it should launch a process that executes a query against redshift. This process is lightweight but may be long-running, making the current lambda infra possibly inappropriate. 

- [x] ECS + fargate infrastructure
- [x] Docker image for pulling from sqs and running queries and writing results to s3
- [x] Docker image schedules file conversion batch job

Issue #193, opened 2019-02-13
Updating API to generate redshift query with bundle uuids


brianraymor:
The current API for the matrix service accepts a list of bundle fqids. We need to preserve that API while swapping out the backend and implementing metadata delivery. So, the new "driver" function needs to accept and queue a list of bundle fqids. A worker translates that into a SQL statement that selects expression data and metadata for those bundles.

- [x] Format queries and store in s3
- [x] Add references to queries in s3 to sqs queue

PR #207, opened 2019-03-01
Redshift query enabling

parthshahva:
This PR is tied to #193 and #194. This PR introduces api changes and ECS infrastructure to run queries on the redshift clusters. It also deletes all references to the old implementation including caching. The ECS infrastructure uses a docker image that packages a query runner and has a templated concurrency. The driver is now responsible for formatting and storing redshift queries in S3. The driver then adds a reference to each query to an sqs queue. The ECS query runner long polls this queue and runs redshift queries that unload the data to the results bucket. After all expected queries have completed, the query runner invokes a matrix conversion job that runs in batch (changes required for this part will land in another PR).

Relevant terraform changes in addition to terraform changes in this PR landed in #201  and https://github.com/HumanCellAtlas/matrix-service/pull/199


config/matrix-api.yml
```diff
@@ -92,8 +91,8 @@ paths:
       summary: "Retrieve the status, and result if available, of a matrix query request."
       description: "Returns the status, and URI of the result if available, of a matrix query request. The status of a
         request can be 'Complete', 'In Progress' or 'Failed'. If the request is 'Complete', the matrix_location field
-        will be populated with an S3 location for 'zarr' formats, and a URL for all other formats. If the request
-        'Failed', the error message will be available in the message field."
+        will be populated with a URL for all other formats. If the request 'Failed',
```
calvinnhieu: *will be populated with a URL to download the matrix.

matrix/common/aws/batch_handler.py
```diff
@@ -11,42 +11,46 @@
...
+                             request_id,
                              format])
 
         is_compressed = format == MatrixFormat.CSV.value or format == MatrixFormat.MTX.value
-        source_zarr_path = f"s3://{self.s3_results_bucket}/{self.request_hash}.zarr"
-        target_path = f"s3://{self.s3_results_bucket}/{self.request_hash}.{format}" + (".zip" if is_compressed else "")
+        source_expression_manifest = f"s3://{self.s3_results_bucket}/{request_id}/expressionmanifest"
+        source_cell_manifest = f"s3://{self.s3_results_bucket}/{request_id}/cell_metadatamanifest"
+        source_gene_manifest = f"s3://{self.s3_results_bucket}/{request_id}/gene_metadatamanifest"
+        target_path = f"s3://{self.s3_results_bucket}/{request_id}.{format}" + (".zip" if is_compressed else "")
```
calvinnhieu: suggest: '_' before "manifest" in prefix?

matrix/common/aws/dynamo_handler.py
```diff
@@ -93,118 +66,95 @@ def _get_dynamo_table_resource_from_enum(self, dynamo_table: DynamoTable):
...
                 StateTableField.EXPECTED_DRIVER_EXECUTIONS.value: 1,
                 StateTableField.COMPLETED_DRIVER_EXECUTIONS.value: 0,
-                StateTableField.EXPECTED_MAPPER_EXECUTIONS.value: num_mappers,
-                StateTableField.COMPLETED_MAPPER_EXECUTIONS.value: 0,
-                StateTableField.EXPECTED_WORKER_EXECUTIONS.value: 0,
-                StateTableField.COMPLETED_WORKER_EXECUTIONS.value: 0,
-                StateTableField.EXPECTED_REDUCER_EXECUTIONS.value: 1,
-                StateTableField.COMPLETED_REDUCER_EXECUTIONS.value: 0,
-                StateTableField.EXPECTED_CONVERTER_EXECUTIONS.value: 0 if format == MatrixFormat.ZARR.value else 1,
+                StateTableField.EXPECTED_QUERY_EXECUTIONS.value: 3,
```
calvinnhieu: Why 3?

matrix/common/aws/dynamo_handler.py
```diff
@@ -93,118 +66,95 @@ def _get_dynamo_table_resource_from_enum(self, dynamo_table: DynamoTable):
...
-    def create_output_table_entry(self, request_hash: str, num_bundles: int, format: str):
+    def create_output_table_entry(self, request_id: str, num_bundles: int, format: str):
         """
         Put a new item in the DynamoDB Table responsible for counting output rows
 
-        :param request_hash: UUID identifying a filter merge job request.
+        :param request_id: UUID identifying a matrix service request.
         :param num_bundles: the number of bundles in the request.
-        :param format: expected file format for filter merge job request.
+        :param format: expected file format for matrix service request.
```
calvinnhieu: thanks :+1:

matrix/common/logging.py
```diff
@@ -21,7 +21,7 @@ def get_logger(name: str):
         return logger
 
     @staticmethod
-    def set_correlation_id(logger: logging.Logger, id_name: str="REQUEST_HASH", value: str=None):
+    def set_correlation_id(logger: logging.Logger, id_name: str="REQUEST_ID", value: str=None):
```
calvinnhieu: nice catch

matrix/common/request/request_tracker.py
```diff
@@ -42,8 +37,8 @@ def __init__(self, request_hash: str):
     @property
     def is_initialized(self) -> bool:
         try:
-            self.dynamo_handler.get_table_item(DynamoTable.OUTPUT_TABLE,
-                                               request_hash=self.request_hash)
+            self.dynamo_handler.get_table_item(DynamoTable.STATE_TABLE,
+                                               request_id=self.request_id)
```
calvinnhieu: Was this arbitrary? Just wondering

calvinnhieu: Ah, request ID is in this table now

matrix/docker/query_runner.py
```diff
@@ -1,9 +1,81 @@
 """Script to pull from sqs and run redshift queries. Will be dockerized."""
```
calvinnhieu: Update comment?

matrix/docker/query_runner.py
```diff
@@ -1,9 +1,81 @@
...
+import json
+import os
 
+from matrix.common.aws.sqs_handler import SQSHandler
+from matrix.common.aws.s3_handler import S3Handler
+from matrix.common.aws.batch_handler import BatchHandler
```
calvinnhieu: nit: alphabetical

matrix/docker/query_runner.py
```diff
@@ -1,9 +1,81 @@
...
+    def query_job_q_url(self):
+        return self.matrix_infra_config.query_job_q_url
+
+    @property
+    def query_job_deadletter_q_url(self):
+        return self.matrix_infra_config.query_job_deadletter_q_url
+
+    def run(self, max_loops=None):
+        loops = 0
+        while True and (max_loops is None or loops < max_loops):
```
calvinnhieu: Is the first `True` for anything?

matrix/common/config.py
```diff
@@ -0,0 +1,13 @@
+from dcplib.config import Config
```
calvinnhieu: I'm assuming this is a wrapper around secrets manager?

config/matrix-api.yml
parthshahva: Thanks!

matrix/common/aws/batch_handler.py
parthshahva: I think I can add '_' to the unload queries. 'Manifest' is added automatically to the key by redshift. Will investigate

matrix/docker/query_runner.py
```diff
@@ -1,9 +1,81 @@
...
+
+    @property
+    def query_job_q_url(self):
+        return self.matrix_infra_config.query_job_q_url
+
+    @property
+    def query_job_deadletter_q_url(self):
+        return self.matrix_infra_config.query_job_deadletter_q_url
+
+    def run(self, max_loops=None):
```
calvinnhieu: Example use case for `max_loops`?

calvinnhieu: suggest: rename and break into smaller functions for testing and clarity

matrix/common/aws/dynamo_handler.py
parthshahva: Lol yes you will. Each matrix service request has 3 queries that need to run currently. The current design has a 1:1 mapping between a query and sqs message. The ecs query runner will pull and run 3 individual queries before scheduling a matrix conversion after the last query finishes execution

parthshahva: This allows us to bump up the concurrency of the ecs query runner in terraform without worrying about any unintended consequences on when conversion is scheduled. This also allows us to easily update the driver to add different queries down the road in cases of filtering, etc

matrix/common/request/request_tracker.py
parthshahva: Yup request ID is key of both state and cache table

matrix/docker/query_runner.py
parthshahva: Hmm update to what? This is exactly what the query runner is doing

matrix/docker/query_runner.py
parthshahva: will fix

matrix/docker/query_runner.py
parthshahva: Yea, this loop will execute forever if no max_loops param is passed in. Essentially this means that each ecs query runner long polls sqs every 15 seconds into eternity.  The max_loops is only set from the test currently

parthshahva: Ohhhh, I see what you are saying. You're right in the True is doing nothing haha. Will remove

matrix/common/config.py
parthshahva: Correct

matrix/docker/query_runner.py
parthshahva: Currently only used by test_run in test_query_runner so that the tests don't run forever. Let me know if you have a better way to handle that

parthshahva: Hmmm, I could remove out some of the sqs message parsing. I think the function calls are clean though and tell the story of the run function

matrix/common/aws/dynamo_handler.py
calvinnhieu: Cool. What currently are the 3 queries? Is there a design doc I can refer to?

matrix/docker/query_runner.py
calvinnhieu: Sounds good. Yea you're right, it's just the parsing and logging that looks clunky. Np

matrix/docker/query_runner.py
calvinnhieu: Thanks for the explanation regardless!

matrix/common/aws/dynamo_handler.py
parthshahva: There is no design doc currently. The 3 queries are defined in driver.py in this PR

parthshahva: I thought you guys already put together the design doc when prototyping redshift (from which the individual units of work were extracted) or are you asking for something different?

calvinnhieu: Just looking to understand what the 3 queries are and what are the potential queries we'd add in the future

matrix/docker/query_runner.py
calvinnhieu: Read "Will be dockerized" as a TODO, nevermind

tests/unit/common/aws/test_sqs_handler.py
```diff
@@ -0,0 +1,45 @@
+import json
+
+from matrix.common.aws.sqs_handler import SQSHandler
+from tests.unit import MatrixTestCaseUsingMockAWS
+
+
+class TestSQSHandler(MatrixTestCaseUsingMockAWS):
```
calvinnhieu: Consider adding a test case to test `MatrixException`

tests/unit/common/aws/test_sqs_handler.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def test_add_message_to_queue(self):
+        payload = {'test_key': "test_value"}
+        self.sqs_handler.add_message_to_queue("test_query_job_q_name", payload)
+
+        messages = self.sqs.meta.client.receive_message(QueueUrl="test_query_job_q_name")
+        message_body = json.loads(messages['Messages'][0]['Body'])
+        self.assertEqual(message_body['test_key'], "test_value")
+
+    def test_retrieve_messages_from_queue__returns_None_when_no_messages_found(self):
```
calvinnhieu: *`test_receieve...`

tests/unit/docker/test_query_runner.py
```diff
@@ -0,0 +1,77 @@
+from unittest import mock
+import uuid
+
+from matrix.docker.query_runner import QueryRunner
+from matrix.common.aws.sqs_handler import SQSHandler
+from tests.unit import MatrixTestCaseUsingMockAWS
+from matrix.common.request.request_tracker import Subtask
+
+
+class TestQueryRunner(MatrixTestCaseUsingMockAWS):
```
calvinnhieu: Test exception handling case

matrix/common/aws/batch_handler.py
mckinsel: I think it just sticks `manifest` on the end of whatever's in the `TO` clause.

mckinsel: Do want to start to handle the user-requested name here?

parthshahva: Yea, we'd probably want to pass that in from the api to driver and then store that user-requested-name in the output table. This can then be used in the batch handler when scheduling conversion job with target path. @mckinsel is there a ticket tracking this already?

brianraymor: We decided against allowing a user requested filename in preference to that being managed by the data browser.

matrix/common/aws/dynamo_handler.py
mckinsel: I'm guessing you mean expression, cell metadata, and gene metadata?

calvinnhieu: Yea. Sorry, I found it later in the PR

matrix/lambdas/daemons/driver.py
```diff
@@ -1,35 +1,90 @@
...
+feature_query_template = """
+    UNLOAD ($$SELECT *
+    FROM feature
+    WHERE feature.isgene$$)
+    to 's3://{0}/{1}/gene_metadata'
+    IAM_ROLE '{2}'
+    GZIP
+    ALLOWOVERWRITE
+    MANIFEST VERBOSE;
+"""
```
mckinsel: I think this is fine because it's a very fast query to run, but this is constant for every request.

matrix/lambdas/daemons/driver.py
```diff
@@ -1,35 +1,90 @@
...
+    LEFT OUTER JOIN feature on (expression.featurekey = feature.featurekey)
+    INNER JOIN cell on (expression.cellkey = cell.cellkey)
+    INNER JOIN analysis on (cell.analysiskey = analysis.analysiskey)
+    WHERE feature.isgene
+    AND expression.exprtype = 'Count'
+    AND analysis.bundle_fqid IN {3}$$)
+    TO 's3://{0}/{1}/expression'
+    IAM_ROLE '{2}'
+    GZIP
+    ALLOWOVERWRITE
```
mckinsel: I think we should remove `ALLOWOVERWRITE` from these. If the files already exist, something has gone wrong.

matrix/common/aws/redshift_handler.py
```diff
@@ -0,0 +1,22 @@
...
+        self.redshift_config = MatrixRedshiftConfig()
+
+    @property
+    def database_uri(self):
+        return self.redshift_config.database_uri
+
+    def run_query(self, query):
+        conn = pg.connect(self.database_uri)
+        cursor = conn.cursor()
+        cursor.execute(query)
```
mckinsel: Do you need to `commit` an `UNLOAD`?

matrix/lambdas/daemons/driver.py
parthshahva: Yea we can optimize later

matrix/lambdas/daemons/driver.py
parthshahva: Done

matrix/common/aws/redshift_handler.py
parthshahva: No I don't believe so. Has been working without any issues

PR #216, opened 2019-03-06
Add definition of next version of API

mckinsel:


config/matrix-api-v1.yml
```diff
@@ -0,0 +1,299 @@
...
+                  - $ref: '#/components/schemas/CategoricalFieldDetail'
+        '404':
+          description: 'Field not found.'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MatrixErrorResponse'
+components:
+  schemas:
+    FilterOperation:
```
brianraymor: I was reviewing Genevieve's [UX Summary](https://docs.google.com/document/d/1cZGlyXzzFCRuAJqcsh58VHQD1apjzAULa_7xQSL60oE/edit#) to help validate the API - avoiding cases that are more query API specific, would we want to support something like _For all pancreas cells from women age 40-50, return all cells where at least 500 genes have non-zero values, **or** total reads > X, etc_.

GDC supports a [broader range](https://docs.gdc.cancer.gov/API/Users_Guide/Search_and_Retrieval/#filters-specifying-the-query) of operators - including  _and_ and _or_. Is the intent to keep operations minimal with _AND_ as the default until further feedback is received?

config/matrix-api-v1.yml
```diff
@@ -0,0 +1,299 @@
...
+                  - $ref: '#/components/schemas/CategoricalFieldDetail'
+        '404':
+          description: 'Filter not found.'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MatrixErrorResponse'
+  /fields:
+    get:
+      summary: Get a lsit of available metadata fields
```
brianraymor: _lsit_ -> _list_

config/matrix-api-v1.yml
```diff
@@ -0,0 +1,299 @@
...
+          description: 'Successfully return field list.'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MatrixFields'
+  /fields/{field_name}:
+    get:
+      summary: 'Get a description of a metadata field'
+      description: >
+        Get detailed information about a an available metadata field.
```
brianraymor: _about ~a~ an availabile_

config/matrix-api-v1.yml
```diff
@@ -0,0 +1,299 @@
...
+    MatrixFilters:
+      description: 'Accepted filter names for a matrix request.'
+      type: array
+      items:
+        type: string
+    MatrixFeature:
+      description: >
+        Feature type to include in the output matrix. For example, 
+        genes vs. transcripts.
+      type: string
```
brianraymor: Could this be an enumeration? What additional _feature types_ would we expect in the future?

config/matrix-api-v1.yml
```diff
@@ -0,0 +1,299 @@
...
+      enum:
+        - '='
+        - '!='
+        - '>'
+        - '<'
+        - 'in'
+    MatrixFormat:
+      description: 'Format for the output matrix.'
+      type: string
```
brianraymor: This was previously an enumeration.

mckinsel: Do you think it should be an enumeration defined in the API or rather described the response to a `/formats` endpoint?

config/matrix-api-v1.yml
mckinsel: It could be. Right now it's just `gene` or `transcript`, but I think you could have `protein` or some non-genic annotation of a region.

config/matrix-api-v1.yml
mckinsel: I had thought to start with _AND_. If we add _OR_, then we need to encode groupings too. That said, if it's already in the UX docs, we probably should just bite the bullet now.

config/matrix-api-v1.md
```diff
@@ -0,0 +1,193 @@
...
+expression matrix along with associated metadata in the format of their choice.
+
+The first version of the matrix service's API focused on integration with the HCA Data Browser. So,
+it required that requests for expression data be specified using "bundle ids". "Bundles" are a
+concept internal to the HCA's data storage system. Since the data browser is closely integrated with
+the storage system, bundle ids were a natural way to specify requests. However, bundles ids make
+less sense for external clients. The focus of this API release is to enable requests for expression
+data specified in terms preferred by a scientist who is unfamiliar with the internals of the HCA's
+data platform.
+
```
theathorn: The Data Browser team (+ @briandoconnor) request that we retain the "URL to a manifest" API long term and in the future add a "reference to a Collection ID" API. These two interfaces allow the Data Browser to pass specific versioned bundles/files to the Matrix Service so that users can reproduce results from e.g. cited data or personal collections or data releases. These APIs should be in addition to this "filter query" interface.

mckinsel:
This was implemented in #285

PR #217, opened 2019-03-09
Create endpoint to subscribe to DSS notifications

calvinnhieu:
New API endpoint to handle DSS notifications for bundle events: `/v0/dss/notifications`
- refactors and reuses [loader scripts](https://github.com/HumanCellAtlas/matrix-service/pull/200) into `matrix/` code directory
- handles `CREATE`, `DELETE` and `TOMBSTONE` events
    - on `CREATE`: update existing tables via [table merge operation](https://docs.aws.amazon.com/redshift/latest/dg/merge-replacing-existing-rows.html)
    - on `DELETE` or `TOMBSTONE`: remove row from `analysis` table given `bundle_fqid`
- deployed on AWS Lambda

Due to the refactor of loader scripts, this PR has absorbed and includes all changes from https://github.com/HumanCellAtlas/matrix-service/pull/219

__Note__: Large number of files affected and large number of additions due to inclusion of test SS2 bundle in `tests/functional/res/etl/bundle/` for Transformer tests


chalice/.chalice/config.json
```diff
@@ -54,7 +58,9 @@
         "LAMBDA_REDUCER_FUNCTION_NAME": "dcp-matrix-service-reducer-predev",
         "DYNAMO_STATE_TABLE_NAME": "dcp-matrix-service-state-table-predev",
         "DYNAMO_OUTPUT_TABLE_NAME": "dcp-matrix-service-output-table-predev",
-        "S3_RESULTS_BUCKET": "dcp-matrix-service-results-predev"
+        "S3_RESULTS_BUCKET": "dcp-matrix-service-results-predev",
+        "MATRIX_VERSION": "56cf1ad",
+        "DYNAMO_CACHE_TABLE_NAME": "dcp-matrix-service-cache-table-predev"
```
parthshahva: How did the cache table make a reappearance? I thought I'd seen it for the last time lol

config/environment
```diff
@@ -17,10 +17,13 @@ fi
 
 AWS_DEFAULT_REGION='us-east-1'
 LAMBDA_DRIVER_FUNCTION_NAME="dcp-matrix-service-driver-${DEPLOYMENT_STAGE}"
+LAMBDA_NOTIFICATIONS_FUNCTION_NAME="dcp-matrix-service-notifications-${DEPLOYMENT_STAGE}"
 DYNAMO_STATE_TABLE_NAME="dcp-matrix-service-state-table-${DEPLOYMENT_STAGE}"
 DYNAMO_OUTPUT_TABLE_NAME="dcp-matrix-service-output-table-${DEPLOYMENT_STAGE}"
 MATRIX_RESULTS_BUCKET="dcp-matrix-service-results-${DEPLOYMENT_STAGE}"
 MATRIX_QUERY_BUCKET="dcp-matrix-service-queries-${DEPLOYMENT_STAGE}"
+MATRIX_PRELOAD_BUCKET="dcp-matrix-service-preload-${DEPLOYMENT_STAGE}"
+MATRIX_REDSHIFT_IAM_ROLE_ARN="arn:aws:iam::${ACCOUNT_ID}:role/matrix-service-redshift-${DEPLOYMENT_STAGE}"
```
parthshahva: Why does this need to be in the environment file? I would think it just needs to exist in terraform?

config/matrix-api.yml
```diff
@@ -125,4 +141,21 @@ paths:
             type: array
             items:
               $ref: '#/definitions/MatrixFormat'
-
+  /dss/notifications:
+    post:
+      summary: "Handles DSS notifications such as creates, updates and deletes."
+      description: "Handles DSS notifications such as bundle creates, updates and deletes and propagates changes to
```
parthshahva: Seems odd to make a reference to redshift (implementation detail) in the API spec?

daemons/driver/requirements.txt
```diff
@@ -1,3 +1,3 @@
 requests==2.20.0
-dcplib==1.4.0
+dcplib==1.5.1
```
parthshahva: We should consider using the root requirements.txt rather than individual requirements.txt files for these daemons. Or is this done to limit package size?

daemons/notifications/app.py
```diff
@@ -0,0 +1,8 @@
+from matrix.lambdas.daemons.notifications import NotificationsHandler
+
+
+def notifications_handler(event, context):
+    assert ('bundle_uuid' in event and 'bundle_version' in event and 'event_type' in event)
+
+    notifications = NotificationsHandler(event['bundle_uuid'], event['bundle_version'], event['event_type'])
```
parthshahva: nit: Since this only ever handles 1 notification should be it called NotificationHandler?

matrix/lambdas/api/core.py
```diff
@@ -154,3 +154,21 @@ def get_matrix(request_id: str):
...
+    bundle_version = body['match']['bundle_version']
+    subscription_id = body['subscription_id']
+    event_type = body['event_type']
+
+    payload = {
+        'bundle_uuid': bundle_uuid,
+        'bundle_version': bundle_version,
+        'event_type': event_type,
+    }
+    lambda_handler.invoke(LambdaName.NOTIFICATIONS, payload)
```
parthshahva: lets consider putting sqs here instead of directly invoking the lambda. There are many reasons to do this 1) Lambda api fails intermittently. Good to have lambda retry driven by sqs 2) control processing rate of notifications 3) track notifications received and failed potentially in a deadletter queue

matrix/common/aws/redshift_handler.py
```diff
@@ -1,12 +1,40 @@
 import psycopg2 as pg
+import typing
+from enum import Enum
 
 from matrix.common.config import MatrixRedshiftConfig
 
 
+class TableName(Enum):
```
parthshahva: The cleanliness of this file makes me very happy

matrix/common/aws/redshift_handler.py
```diff
@@ -19,4 +47,13 @@ def run_query(self, query):
         conn = pg.connect(self.database_uri)
         cursor = conn.cursor()
         cursor.execute(query)
+        conn.commit()
+        conn.close()
+
+    def transaction(self, queries: typing.List[str]):
```
parthshahva: can we just replace run_query with transaction in the query_runnner (invoked with a list of one query)? This way we can delete one of these functions

matrix/common/constants.py
```diff
@@ -15,3 +15,117 @@ class MatrixRequestStatus(Enum):
     COMPLETE = "Complete"
     IN_PROGRESS = "In Progress"
     FAILED = "Failed"
+
+
+CREATE_QUERY_TEMPLATE = {
```
parthshahva: Do we need to manage a migration framework? Or will we just rebuild the entire database anytime we need to make changes to this schema?

matrix/common/etl/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+        query=query,
+        transformer=transformer_cb,
+        finalizer=finalizer_cb,
+        max_workers=max_workers
+    )
+
+
+def transform_bundle(bundle_uuid: str, bundle_version: str, bundle_path: str, extractor: DSSExtractor):
+    """
+    Transforms a downloaded bundle into PSV rows - per bundle transformer passed to ETL.
```
parthshahva: What does PSV stand for?

matrix/common/etl/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+    logger.info(f"ETL: All transformations complete.")
+    _upload_to_s3(os.path.join(extractor.sd, MetadataToPsvTransformer.OUTPUT_DIRNAME), job_id)
+    _populate_all_tables(job_id, temp=True)
+
+
+def _upload_to_s3(output_dir, job_id: str):
+    """
+    Uploads to S3 all files in OUTPUT_DIR.
+    """
+    for name in os.listdir(output_dir):
```
parthshahva: We should parallelize the upload to s3. I did this in the hca cli and saw 5x improvement

matrix/common/etl/__init__.py
```diff
@@ -0,0 +1,219 @@
...
+
+    logger.info(f"ETL: Populating Redshift tables. Committing transaction.")
+    try:
+        redshift.transaction(transaction)
+    except psycopg2.Error as e:
+        # TODO: put this in a DLQ
+        logger.error("Failed to populate Redshift tables. Rolling back.", e)
+
+
+def get_dss_client(deployment_stage: str):
```
parthshahva: can we move this to a utils maybe? This same function will also exist in the functional tests

matrix/lambdas/daemons/notifications.py
```diff
@@ -0,0 +1,53 @@
...
+                transformer_cb=transform_bundle,
+                finalizer_cb=finalizer_update,
+                staging_directory=os.path.abspath("/tmp"),
+                deployment_stage=os.environ['DEPLOYMENT_STAGE'],
+                max_workers=16)
+
+        logger.info(f"Done processing DSS notification for {self.bundle_uuid}.{self.bundle_version}")
+
+    def remove_bundle(self):
+        delete_query = f"DELETE FROM analysis WHERE bundle_fqid='{self.bundle_uuid}.{self.bundle_version}'"
```
parthshahva: Is this really the only table we need to delete an entry from on a delete notification? What about cell/expression?

matrix/lambdas/daemons/notifications.py
```diff
@@ -0,0 +1,53 @@
...
+        self.redshift = RedshiftHandler()
+
+    def run(self):
+        if self.event_type == 'CREATE':
+            self.update_bundle()
+        elif self.event_type == 'DELETE' or self.event_type == 'TOMBSTONE':
+            self.remove_bundle()
+
+    def update_bundle(self):
+        query = {
```
parthshahva: Do we need to check if this is an existing bundle first and do a check on versions? Do we need to delete before inserting? Does this overwrite each time?

terraform/modules/matrix-service/lambdas/notifications_lambda.tf
```diff
@@ -0,0 +1,99 @@
...
+  "Statement": [
+    {
+      "Sid": "LogsPolicy",
+      "Effect": "Allow",
+      "Resource": [
+        "arn:aws:logs:${var.aws_region}:${var.account_id}:log-group:/aws/lambda/dcp-matrix-service-notifications-${var.deployment_stage}",
+        "arn:aws:logs:${var.aws_region}:${var.account_id}:log-group:/aws/lambda/dcp-matrix-service-notifications-${var.deployment_stage}:*:*"
+      ],
+      "Action": [
+        "logs:CreateLogGroup",
```
parthshahva: Does this work? I thought logs:CreateLogGroup needs * resource path?

terraform/modules/matrix-service/lambdas/notifications_lambda.tf
```diff
@@ -0,0 +1,99 @@
...
+
+resource "aws_lambda_function" "matrix_service_notifications_lambda" {
+  function_name    = "dcp-matrix-service-notifications-${var.deployment_stage}"
+  s3_bucket        = "${var.deployment_bucket_id}"
+  s3_key           = "notifications_daemon.zip"
+  role             = "${aws_iam_role.matrix_service_notifications_lambda.arn}"
+  handler          = "app.notifications_handler"
+  runtime          = "python3.6"
+  timeout          = 900
+  memory_size      = 3008
```
parthshahva: Is this the max memory size? Why was this chosen?

terraform/modules/matrix-service/redshift/s3.tf
```diff
@@ -1,8 +1,8 @@
-resource "aws_s3_bucket" "matrix_service_redshift_preload" {
-  bucket = "dcp-matrix-service-redshift-preload-${var.deployment_stage}"
+resource "aws_s3_bucket" "matrix_service_preload" {
+  bucket = "dcp-matrix-service-preload-${var.deployment_stage}"
 
   lifecycle_rule {
-    id      = "matrix_service_redshift_preload_expiration"
+    id      = "matrix_service_preload_expiration"
 
     expiration {
```
parthshahva: Why do we have an expiration of 7 days?

scripts/redshift/load_redshift.py
```diff
@@ -0,0 +1,44 @@
...
+
+if __name__ == '__main__':
+    # All SS2 analysis bundles
+    query = {
+        "query": {
+            "bool": {
+                "must": [
+                    {
+                        "match": {
+                            "files.library_preparation_protocol_json.library_construction_approach.ontology": "EFO:0008931"
```
parthshahva: I'm guessing this will need to change with the recent metadata change?

scripts/redshift/load_redshift.py
```diff
@@ -0,0 +1,44 @@
...
+                        }
+                    },
+                    {
+                        "match": {
+                            "files.donor_organism_json.biomaterial_core.ncbi_taxon_id": 9606
```
parthshahva: Can we leave a comment as to why this query has a value of 9606?

matrix/common/etl/transformers/cell_expression.py
```diff
@@ -136,14 +139,23 @@ def _parse_10x_bundle(self, bundle_dir):
...
-            # decide if this is a lucky cell
-            is_lucky = str(ord(cell_key[-1]) % 5 == 0)
+            cell_to_barcode[cell_key] = barcode
 
-            expression_line = '|'.join(
-                [cell_key,
-                 gene,
-                 "Count",
-                 str(v)]) + '\n'
+            if cell_key not in cell_gene_counts:
```
parthshahva: This code is a little difficult to understand but fine for revisiting it at a later time

config/matrix-api.yml
```diff
@@ -56,6 +56,22 @@ definitions:
     properties:
       message:
         type: string
+  DssNotification:
```
parthshahva: As mentioned in slack, I was surprised to not see the logic for creating subscriptions in this PR. I think we should be deleting and recreating the notifications as part of the deployment flow with the queries explicitly defined in the code? Perhaps have a NotificationRegister (terrible name, pick something else) class where this is defined

matrix/common/etl/__init__.py
```diff
@@ -76,7 +76,10 @@ def transform_bundle(bundle_uuid: str, bundle_version: str, bundle_path: str, ex
     ]
 
     for transformer in transformers:
-        transformer.transform(bundle_path)
+        try:
+            transformer.transform(bundle_path)
+        except Exception as e:
+            logger.error(f"Failed to transform bundle {bundle_uuid}.{bundle_version}.", e)
```
parthshahva: logging an error doesn't seem sufficient

chalice/.chalice/config.json
calvinnhieu: Ah, that was from me testing the Cache Invalidation work on `predev`

config/environment
calvinnhieu: Redshift [COPY command](https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html#r_COPY-syntax-overview) requires an argument that specifies an IAM role to use.

config/matrix-api.yml
calvinnhieu: Ah, yes revising

config/matrix-api.yml
calvinnhieu: +1 on version controlling the query and recreating it per deployment. Work captured [here](https://app.zenhub.com/workspaces/dcp-backlogs-5ac7bcf9465cb172b77760d9/issues/humancellatlas/matrix-service/224)

daemons/driver/requirements.txt
calvinnhieu: Yup to limit package size. I'm blanking on which part of the deployment actually uses the root `requirements.txt` - can you remind me?

daemons/notifications/app.py
calvinnhieu: Sure, renaming

matrix/common/aws/redshift_handler.py
calvinnhieu: Yup, didn't want to introduce any unnecessary regressions but I'll go ahead and make that change

matrix/common/constants.py
calvinnhieu: Maybe we can revisit this question after getting performance/load times with `dcplib/ETL` parallelizing the transformers as well. Currently, a download of integration (~10K bundles) completes within an hour, but the single-threaded execution of transformers takes ~7 hours. I'd imagine we'd be below 1 or 2 hours with the full parallelization

calvinnhieu: After a quick search, it looks like [migrating tables to a new schema](https://stackoverflow.com/questions/22548928/how-to-change-table-schema-after-created-in-redshift) may be simple.

matrix/common/etl/__init__.py
calvinnhieu: Pipe separated value

matrix/common/etl/__init__.py
calvinnhieu: Agreed. As we discussed, errors should be propagated and stored in a DLQ. Work captured [here](https://app.zenhub.com/workspaces/dcp-backlogs-5ac7bcf9465cb172b77760d9/issues/humancellatlas/matrix-service/227)

matrix/common/etl/__init__.py
calvinnhieu: The upload is relatively small/quick as we're uploading text files (seconds compared to hours). Granted it's worth double-checking the size of the expression files being uploaded. Will do a quick check on this.

matrix/common/etl/__init__.py
calvinnhieu: Sure, will refactor when it's reintroduced in the tests

matrix/common/etl/transformers/cell_expression.py
calvinnhieu: Agreed, the actual metadata transformations could use a little clean up, especially as they're only going to get more complex with updates to metadata schema.

matrix/lambdas/api/core.py
calvinnhieu: Agreed: https://app.zenhub.com/workspaces/dcp-backlogs-5ac7bcf9465cb172b77760d9/issues/humancellatlas/matrix-service/227

matrix/lambdas/daemons/notifications.py
calvinnhieu: This is the simplest delete implementation I could think of given that our inputs for the first iteration are just `bundle_fqid`. The issue with deleting rows from more tables such as cell and expression - namely cell - was that we had a weak assumption that our `cellkey`s are unique across bundles. However, I believe that assumption is actually true with the way we're computing `cellkey` now, ignoring the `provenance.document_id`.

Another issue with deleting more rows is with tables that have a many:X relationship with bundles: project, contributor, specimen, etc, in which the update logic becomes more complex (if a single bundle updates or changes its project, how should that project entry be updated considering its relation to other bundles/cells?)

terraform/modules/matrix-service/lambdas/notifications_lambda.tf
calvinnhieu: Yup, I'm getting logs in CloudWatch

matrix/lambdas/daemons/notifications.py
calvinnhieu: Good catch. I didn't implement a versions check since we aren't expecting updates to come through the DSS, but it would be simple enough to implement if we were expecting for it to be used. Prefer not to if we didn't have to to avoid issuing an unnecessary query.

Nothing would happen if the bundle doesn't exist, but it would be cleaner to short circuit this if no bundle existed. I'll check the DSS API to see what's the quickest way to check for the existence of a bundle.

In `finalizer_update`, it runs through [this process](https://docs.aws.amazon.com/redshift/latest/dg/merge-replacing-existing-rows.html):
1. transforms the bundle into temporary tables
2. replace rows in real tables with rows in temporary tables

Yes, it will delete before inserting and overwrites conflicting primary keys every time.

terraform/modules/matrix-service/lambdas/notifications_lambda.tf
calvinnhieu: Yup [max memory size](https://docs.aws.amazon.com/lambda/latest/dg/limits.html). Unsure how large 10X bundles can get

terraform/modules/matrix-service/redshift/s3.tf
calvinnhieu: Do you suggest it should be lower or higher or have none? The staged PSVs from a notification has little use beyond debugging purposes. A better argument can be made for staged PSVs from a full load. Maybe a full load should use a different bucket than notifications where the former has a longer expiration than the latter (to preserve a sort of a snapshot to perform expedited reloads)?

scripts/redshift/load_redshift.py
calvinnhieu: Yes

terraform/modules/matrix-service/redshift/s3.tf
calvinnhieu: Alternatively, it could be a single bucket with a longer (30 day?) expiration

daemons/driver/requirements.txt
parthshahva: Actually don't think any part of the deployments uses the root requirements.txt at the moment

matrix/common/constants.py
parthshahva: Cool I'm good with not worrying about this at the moment

matrix/common/etl/__init__.py
parthshahva: Its worth also checking the # of files uploaded. Even if they are small, there tend to be huge wins in parallelization

matrix/lambdas/daemons/notifications.py
parthshahva: As long as the queries in the driver won't return duplicate results, I'm good with this

matrix/lambdas/daemons/notifications.py
parthshahva: Makes sense, especially considering we are a long way away from updated bundles

terraform/modules/matrix-service/lambdas/notifications_lambda.tf
parthshahva: Makes sense, we can tune later

terraform/modules/matrix-service/redshift/s3.tf
parthshahva: Hmm, yea, I prefer 30 days if we are keeping something around for debugging purposes. Treat it like logs in a way

scripts/redshift/load_redshift.py
calvinnhieu: Yup, for context, 9606 is the ID for "Homo Sapiens"

PR #219, opened 2019-03-11
Updates to schema and load scripts

mckinsel:
A few cleanup items:

1. Fix diststyle and sorting for a couple table.
2. Change the way expression counts are gathered so we don't get duplicates.
3. Change the 10x cellkey calculation to just use bundle id.
4. Expand organ parsing to handle non-organoid cell lines.
5. Switch the donor_organism table to a specimen table to handle donors with multiple organs.

### DIFF
--- a/scripts/redshift/transformers/cell_expression.py
+++ b/scripts/redshift/transformers/cell_expression.py
@@ -76,7 +76,7 @@ class CellExpressionTransformer(MetadataToPsvTransformer):
         cell_lines = ['|'.join([
             cell_key,
             keys["project_key"],
-            keys["donor_key"],
+            keys["specimen_key"],
             keys["library_key"],
             keys["analysis_key"],
             "",
@@ -150,15 +150,12 @@ class CellExpressionTransformer(MetadataToPsvTransformer):
             gene_count = len(gene_count_dict)
             cell_line = '|'.join(
                 [cell_key,
-                 keys["project_key"],
-                 keys["donor_key"],
-                 keys["library_key"],
-                 keys["analysis_key"],
-                 barcode,
-                 str(gene_count)]) + '\n'
-            cell_lines.add(cell_line)
-
-            expression_lines.append(expression_line)
+                keys["project_key"],
+                keys["specimen_key"],
+                keys["library_key"],
+                keys["analysis_key"],
+                cell_to_barcode[cell_key],
+                str(gene_count)]) + '\n'
             cell_lines.add(cell_line)
 
         return cell_lines, expression_lines
@@ -172,21 +169,17 @@ class CellExpressionTransformer(MetadataToPsvTransformer):
         ap_path = p.joinpath("analysis_protocol_0.json")
         ap_key = json.load(open(ap_path))["provenance"]["document_id"]
 
-        donor_paths = list(p.glob("donor_organism_*.json"))
-        if len(donor_paths) > 1:
-            donor_key = ""
-        else:
-            donor_key = json.load(open(donor_paths[0]))["provenance"]["document_id"]
+        specimen_paths = list(p.glob("specimen_from_organism_*.json"))
+        specimen_keys = [json.load(open(p))["provenance"]["document_id"] for p in specimen_paths]
+        specimen_key = sorted(specimen_keys)[0]
 
         library_paths = list(p.glob("library_preparation_protocol_*.json"))
-        if len(library_paths) > 1:
-            library_key = ""
-        else:
-            library_key = json.load(open(library_paths[0]))["provenance"]["document_id"]
+        library_keys = [json.load(open(p))["provenance"]["document_id"] for p in library_paths]
+        library_key = sorted(library_keys)[0]
 
         return {
             "project_key": project_key,
-            "donor_key": donor_key,
+            "specimen_key": specimen_key,
             "library_key": library_key,
             "analysis_key": ap_key
         }
--- a/scripts/redshift/transformers/cell_expression.py
+++ b/scripts/redshift/transformers/cell_expression.py
@@ -53,21 +53,26 @@ class CellExpressionTransformer(MetadataToPsvTransformer):
         with open(isoforms_path) as iso_file:
             reader = csv.DictReader(iso_file, delimiter='\t')
             for row in reader:
-                isoforms_values[row["transcript_id"].split(".")[0]] = {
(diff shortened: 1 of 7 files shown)


scripts/redshift/transformers/cell_expression.py
```diff
@@ -108,48 +113,49 @@ def _parse_10x_bundle(self, bundle_dir):
...
+            cell_to_barcode[cell_key] = barcode
 
-            expression_line = '|'.join(
-                [cell_key,
-                 gene,
-                 "Count",
-                 str(v)]) + '\n'
+            if cell_key not in cell_gene_counts:
+                cell_gene_counts[cell_key] = {}
+            cell_gene_counts[cell_key][gene] = cell_gene_counts[cell_key].get(gene, 0) + v
```
calvinnhieu: This doesn't consolidate the "autosomal" section of X and Y chromosomes does it?

scripts/redshift/transformers/cell_expression.py
```diff
@@ -108,48 +113,49 @@ def _parse_10x_bundle(self, bundle_dir):
...
             gene = genes[i]
-            gene_count = genes_detected.item((0, j))
 
             # Just make up a cell id
             h = hashlib.md5()
-            h.update(keys["project_key"].encode())
-            h.update(keys["donor_key"].encode())
-            h.update(cell_suspension_id.encode())
+            bundle_uuid = pathlib.Path(bundle_dir).parts[-1]
+            h.update(bundle_uuid.encode())
```
calvinnhieu: Oh so `bundle_uuid` (i.e. analysis table rows) have a 1:many relationship with cell and expression rows, and not many:many? This would simplify our DSS subscriptions given cellkey is unique to a bundle

scripts/redshift/transformers/cell_expression.py
```diff
@@ -163,21 +169,17 @@ def _parse_keys(self, bundle_dir):
...
         ap_key = json.load(open(ap_path))["provenance"]["document_id"]
 
-        donor_paths = list(p.glob("donor_organism_*.json"))
-        if len(donor_paths) > 1:
-            donor_key = ""
-        else:
-            donor_key = json.load(open(donor_paths[0]))["provenance"]["document_id"]
+        specimen_paths = list(p.glob("specimen_from_organism_*.json"))
+        specimen_keys = [json.load(open(p))["provenance"]["document_id"] for p in specimen_paths]
+        specimen_key = sorted(specimen_keys)[0]
```
calvinnhieu: Can you explain why we can ignore the other keys?

scripts/redshift/transformers/specimen_library.py
```diff
@@ -119,50 +119,58 @@ def create_ontology_resolver(self, *metadata_dict_lists):
...
             dev_stage = donor_dict.get("development_stage", {}).get("ontology", "").upper()
 
-            donorkeys.append(key)
-            genera.add(genus_species)
             ethnicities.add(ethnicity)
-            diseases.add(disease)
             dev_stages.add(dev_stage)
 
-        # Some cells have multiple donors. Only record the donor information if
+        # Some cells have multiple donors/specimens. Only record the donor information if
```
calvinnhieu: Sorry, can you explain how this is being done again?

calvinnhieu: Is there value in/do we plan on supporting multi-valued fields in the future? ex. list of specimen keys in `cell` table

scripts/redshift/transformers/cell_expression.py
mckinsel: The `gene` strips off everything after the first period, which means things like `PAR_Y` and version get stripped off. So the PAR would get consolidated into a single entry. I think that's what we want. There's a related issue here: https://github.com/HumanCellAtlas/secondary-analysis/issues/547

scripts/redshift/transformers/cell_expression.py
mckinsel: So there's usually only one specimen. But in cases where there's more than one, we want to report as much as we can. For example, there are a bunch of bundles currently in the prod DSS that have four donor organisms and four specimens that end up in a single cell suspension. Each of those donors is different, different ages, sex, etc, but they share a bunch of attributes like genus or disease. The strategy here is to just pick one of the ids and attach all the common attributes to it. There's a similar `sorted(...)[0]` in `SpecimenLibraryTransformer._parse_keys`.

scripts/redshift/transformers/specimen_library.py
mckinsel: I think there is, but I don't know how to implement it.

mckinsel:
Okay sounds good. Just so I'm clear, the plan is to just close this once you merge #217?

calvinnhieu:
Ya let's do that, sounds cleaner - I was saying merge this and then I'll port.

scripts/redshift/transformers/specimen_library.py
calvinnhieu: Sounds good. We could look into it further, but a `LIKE` query on `donorkey` where the value is a CSV may work

scripts/redshift/transformers/cell_expression.py
calvinnhieu: I see. So it's unlikely (or not possible) for a single specimen to show up in another bundle with a different set of specimens? i.e. this `sorted(...)[0]` should uniquely identify a set of specimens/donors?

scripts/redshift/transformers/cell_expression.py
calvinnhieu: Cool!

scripts/redshift/transformers/specimen_library.py
mckinsel: Yeah we'd have to see how that impacts performance. They tell you not to use `LIKE` very much in the [redshift docs](https://docs.aws.amazon.com/redshift/latest/dg/c_designing-queries-best-practices.html).

scripts/redshift/transformers/cell_expression.py
mckinsel: Hmmm yeah that's a good question. I don't think that has happened yet, but I'm assuming it could.

calvinnhieu:
Closing, changes have been moved to https://github.com/HumanCellAtlas/matrix-service/pull/217

Issue #195, opened 2019-02-13
Updates to file conversion


brianraymor:
Previously, the file conversion docker image expected a zarr store that expression values and no metadata. Now, it needs to handle the psv output of a redshift query that includes metadata for cells and genes.

AC:
New docker image that passes functional tests and produces a csv, mtx, and loom along with associated metadata.

- [ ] Convert redshift manifests into csv, loom, mtx formats
- [ ] Bubble up file conversion errors to request table.
- [ ] README included in the zip file that documents the output

brianraymor:
Could also consider:

- [ ] README included in the zip file that documents the output

PR #222, opened 2019-03-12
WIP matrix conversion and passing functional tests

parthshahva:
Things left
- add readme
- log error from converting in request output table item


parthshahva:
Related to #195

Issue #227, opened 2019-03-14
Improve Notifications endpoint error handling with SQS and DLQ


calvinnhieu:
Currently, the Notifications endpoint will log errors to CloudWatch and rollback any Redshift queries upon failure when processing a DSS notification. This solution implements SQS and a DLQ to track failed notifications. 

DoD:
- [ ] Implement an SQS queue between the API Gateway Notifications endpoint and the Notifications lambda
- [ ] Capture and propagate all errors to a DLQ. DLQ entry should include the following:
    - [ ] bundle_fqid
    - [ ] notification event type
    - [ ] subscription_id
    - [ ] (reference to) error/stack trace

PR #242, opened 2019-03-25
put sqs between api and notification lambda

parthshahva:
This is related to #227. It adds sqs/dlq between the matrix service api and the notification lambda. It also adds a functional test that tests the /dss/notification endpoint

### DIFF
--- a/config/iam_policy_templates/matrix-service-api-lambda.json
+++ b/config/iam_policy_templates/matrix-service-api-lambda.json
@@ -23,8 +23,7 @@
         "lambda:InvokeFunction"
       ],
       "Resource": [
-        "arn:aws:lambda:us-east-1:${account_id}:function:dcp-matrix-service-driver-${DEPLOYMENT_STAGE}",
-        "arn:aws:lambda:us-east-1:${account_id}:function:dcp-matrix-service-notification-${DEPLOYMENT_STAGE}"
+        "arn:aws:lambda:us-east-1:${account_id}:function:dcp-matrix-service-driver-${DEPLOYMENT_STAGE}"
       ]
     },
     {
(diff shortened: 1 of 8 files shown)


config/iam_policy_templates/matrix-service-api-lambda.json
```diff
@@ -23,8 +23,7 @@
         "lambda:InvokeFunction"
       ],
       "Resource": [
-        "arn:aws:lambda:us-east-1:${account_id}:function:dcp-matrix-service-driver-${DEPLOYMENT_STAGE}",
-        "arn:aws:lambda:us-east-1:${account_id}:function:dcp-matrix-service-notification-${DEPLOYMENT_STAGE}"
```
calvinnhieu: was this removed on purpose?

calvinnhieu: Oh nevermind I see

terraform/modules/matrix-service/infra/sqs.tf
```diff
@@ -11,3 +11,24 @@ resource "aws_sqs_queue" "query_deadletter_queue" {
   name                      = "dcp-matrix-query-deadletter-queue-${var.deployment_stage}"
   message_retention_seconds = 1209600
 }
+
+resource "aws_sqs_queue" "notification_queue" {
+  name                      = "dcp-matrix-notification-queue-${var.deployment_stage}"
+//  Queue visibility timeout must be larger than (triggered lambda) function timeout
```
calvinnhieu: Humor me - could you explain why this is?

tests/functional/test_matrix_service.py
```diff
@@ -138,6 +141,27 @@ def _post_matrix_service_request(self, bundle_fqids=None, bundle_fqids_url=None,
...
+        data["match"]["bundle_version"] = bundle_version
+
+        response = self._make_request(description="POST NOTIFICATION TO MATRIX SERVICE",
+                                      verb='POST',
+                                      url=f"{self.api_url}/dss/notification",
+                                      expected_status=200,
+                                      data=json.dumps(data),
+                                      headers=self.headers)
+        data = json.loads(response)
+        return data
```
calvinnhieu: Do you think it would be worthwhile to validate the rows exist in Redshift?

terraform/modules/matrix-service/infra/sqs.tf
parthshahva: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html is a great read

tests/functional/test_matrix_service.py
parthshahva: We're sending notifications for one the 5 bundles used in test so it should already be in redshift. Are you suggesting we send a delete notification first, make sure its deleted, then create, and then make sure its there?

PR #252, opened 2019-03-29
fix entrypoint into notification lambda for sqs and add functional test for notification

parthshahva:
fix entrypoint into notification lambda for sqs and add complete functional test for notification processing

### DIFF
--- a/matrix/common/aws/redshift_handler.py
+++ b/matrix/common/aws/redshift_handler.py
@@ -43,10 +43,14 @@ class RedshiftHandler:
     def database_uri(self):
         return self.redshift_config.database_uri
 
-    def transaction(self, queries: typing.List[str]):
+    def transaction(self, queries: typing.List[str], return_results=False):
+        results = []
         conn = pg.connect(self.database_uri)
         cursor = conn.cursor()
         for query in queries:
             cursor.execute(query)
+        if return_results:
+            results = cursor.fetchall()
         conn.commit()
         conn.close()
+        return results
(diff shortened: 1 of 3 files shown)


matrix/common/aws/redshift_handler.py
```diff
@@ -50,3 +50,12 @@ def transaction(self, queries: typing.List[str]):
             cursor.execute(query)
         conn.commit()
         conn.close()
+
+    def run_query_and_return_results(self, query):
+        conn = pg.connect(self.database_uri)
+        cursor = conn.cursor()
+        cursor.execute(query)
+        results = cursor.fetchall()
```
calvinnhieu: Can this be added to `transaction` as an optional param instead?

tests/functional/test_matrix_service.py
```diff
@@ -88,8 +90,26 @@ def test_matrix_service_without_specified_output(self):
...
 
+    @unittest.skipUnless(os.getenv('DEPLOYMENT_STAGE') != "prod",
+                         "Do not want to process fake notifications in production.")
     def test_dss_notification(self):
-        self._post_notification(bundle_fqid=INPUT_BUNDLE_IDS[self.dss_env][0])
+        bundle_fqid = INPUT_BUNDLE_IDS[self.dss_env][0]
+
+        self._post_notification(bundle_fqid=bundle_fqid, event_type="DELETE")
+        WaitFor(self._poll_db_get_analysis_row_count_from_fqid, bundle_fqid)\
+            .to_return_value(0, timeout_seconds=300)
```
calvinnhieu: sugg: `DELETE` and `TOMBSTONE` are trivial updates (remove row) compared to `CREATE` (merge tables) and could use a shorter timeout (60s)

parthshahva: Sounds good

matrix/common/aws/redshift_handler.py
parthshahva: Sounds good

PR #267, opened 2019-04-08
Reduce memory consumption in conversion job

mckinsel:
Previously, the conversion jobs were using an enormous amount of memory when converting all the data in the HCA at once. This reduces that to something that runs on an m4.xlarge and removes the relationship between request size and memory usage.


calvinnhieu:
No changes required to `_to_csv` or `_to_mtx`? Also, bump the docker image version by:
- `make release` in `docker-images/matrix-converter`
- update batch job defn version in `converter_batch.tf`

mckinsel:
No those are coming, so this isn't ready to merge yet. Will bump the version.

parthshahva:
Is 16 GiB (m4.xlarge) the max memory used by conversion jobs or is it a lower amount? We should update https://github.com/HumanCellAtlas/matrix-service/blob/master/terraform/modules/matrix-service/infra/converter_batch.tf#L226 to reflect the new memory and cpu requirements in this PR

parthshahva:
See comments above. This PR still needs a new conversion image version and updates to the terraform config for how much cpu/memory is required. I'm honestly also surprised by the lack of additional test coverage and how large some of these functions have become. It's becoming quite difficult to understand matrix conversions

mckinsel:
I added a functional test that looks at the outputs for each of the three formats. It's too slow to run on every commit, but it compares each file against the redshift outputs directly.

tests/functional/test_conversions.py
```diff
@@ -0,0 +1,290 @@
...
+import loompy
+import s3fs
+import scipy
+
+from matrix.docker.matrix_converter import MatrixConverter
+
+# This is the 2544 pancreas data in a public bucket.
+CELL_MANIFEST = "s3://hca-matrix-conversion-test-data/pancreas/cell_metadata_manifest"
+GENE_MANIFEST = "s3://hca-matrix-conversion-test-data/pancreas/gene_metadata_manifest"
+EXPRESSION_MANIFEST = "s3://hca-matrix-conversion-test-data/pancreas/expression_manifest"
```
calvinnhieu: Could use `dcp-matrix-test-data` bucket

calvinnhieu:
How long does the test take?

Issue #248, opened 2019-03-27
Update GET endpoint to handle failed cases


calvinnhieu:
Currently, the GET endpoint will be stuck In Progress for 1) hanging requests and 2) uncaught errors during Batch conversion (e.g. OOM failed container). Update the GET endpoint to properly report these failures in its response by:

DoD:
- [x] Handle FAILED case before COMPLETE case
- [x] [Re-implement](https://github.com/HumanCellAtlas/matrix-service/pull/206) request timeout (24hr)
- [x] Also check Batch job status to determine request status (store Batch job ID in Output table)

PR #269, opened 2019-04-11
Better error reporting for batch failure and timeouts

parthshahva:
This PR accomplishes ticket #248. 1) Adds back 12 hr timeout on jobs 2) Returns failed request status if the batch job is marked as failed (saves corresponding batch job id in state table) 3) Consolidates method to write strings to dynamo tables and removes method specific to writing errors to output table

### DIFF
--- a/matrix/common/aws/batch_handler.py
+++ b/matrix/common/aws/batch_handler.py
@@ -1,11 +1,11 @@
 import os
 
 import boto3
+from tenacity import retry, stop_after_attempt, wait_fixed
 
 from matrix.common.aws.dynamo_handler import DynamoTable
 from matrix.common.constants import MatrixFormat
 from matrix.common.logging import Logging
-from matrix.common.aws.cloudwatch_handler import CloudwatchHandler, MetricName
 
 logger = Logging.get_logger(__name__)
 
@@ -18,8 +18,8 @@ class BatchHandler:
         self.job_def_arn = os.environ['BATCH_CONVERTER_JOB_DEFINITION_ARN']
 
         self._client = boto3.client("batch", region_name=os.environ['AWS_DEFAULT_REGION'])
-        self._cloudwatch_handler = CloudwatchHandler()
 
+    @retry(reraise=True, wait=wait_fixed(2), stop=stop_after_attempt(5))
     def schedule_matrix_conversion(self, request_id: str, format: str):
         """
         Schedule a matrix conversion job within aws batch infra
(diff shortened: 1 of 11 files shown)


matrix/common/aws/batch_handler.py
```diff
@@ -18,8 +18,8 @@ def __init__(self):
         self.job_def_arn = os.environ['BATCH_CONVERTER_JOB_DEFINITION_ARN']
 
         self._client = boto3.client("batch", region_name=os.environ['AWS_DEFAULT_REGION'])
-        self._cloudwatch_handler = CloudwatchHandler()
 
+    @retry(reraise=True, wait=wait_fixed(2), stop=stop_after_attempt(5))
```
calvinnhieu: What happened before if the enqueuing failed?

matrix/common/aws/dynamo_handler.py
```diff
@@ -161,6 +151,44 @@ def increment_table_field(self, table: DynamoTable, request_id: str, field_enum:
...
+            try:
+                table.update_item(
+                    Key=key_dict,
+                    UpdateExpression=f"SET {field_enum_value} = :n",
+                    ExpressionAttributeValues={":n": field_value}
+                )
+                break
+            except botocore.exceptions.ClientError as exc:
+                if exc.response['Error']['Code'] == "ConditionalCheckFailedException":
+                    pass
```
calvinnhieu: Could this cause an infinite loop? i.e. when would this case run?

matrix/common/aws/dynamo_handler.py
```diff
@@ -88,7 +89,8 @@ def create_state_table_entry(self,
                 StateTableField.COMPLETED_QUERY_EXECUTIONS.value: 0,
                 StateTableField.EXPECTED_CONVERTER_EXECUTIONS.value: 1,
                 StateTableField.COMPLETED_CONVERTER_EXECUTIONS.value: 0,
-                StateTableField.CREATION_DATE.value: date.get_datetime_now(as_string=True)
+                StateTableField.CREATION_DATE.value: date.get_datetime_now(as_string=True),
+                StateTableField.BATCH_JOB_ID.value: "N/A"
```
calvinnhieu: Consider storing "N/A" in `constants.py`

matrix/common/request/request_tracker.py
```diff
@@ -209,8 +245,24 @@ def log_error(self, message: str):
...
+
+    def write_batch_job_id_to_db(self, batch_job_id: str):
+        """
+        Logs the batch job id for matrix conversion to state table
+        """
+        self.dynamo_handler.set_table_field_with_value(DynamoTable.STATE_TABLE,
+                                                       self.request_id,
+                                                       StateTableField.BATCH_JOB_ID,
+                                                       batch_job_id)
+        self.cloudwatch_handler.put_metric_data(
```
calvinnhieu: Does this make more sense here or in `batch_handler.schedule_matrix_conversion`?

matrix/common/aws/batch_handler.py
parthshahva: exception raised that bubbled up to query runner

matrix/common/aws/dynamo_handler.py
parthshahva: yea I'm deleting this code. This was only useful for when multiple lambdas were incrementing completed executions (ie copy and paste leftover)

matrix/common/request/request_tracker.py
parthshahva: changing

matrix/common/aws/dynamo_handler.py
parthshahva: Considered. Leaving as is

PR #281, opened 2019-05-02
log transform errors during ETL

calvinnhieu:
These changes log transformation errors in `logs/errors.txt` for further review. Cell and expression transforms are performed per bundle and will yield a `bundle_uuid.bundle_version` on failure, whereas all other transformers are performed against all bundles and will simply report if the transformer failed.


calvinnhieu:
Updated log to include stack traces; separately log failed bundle fqids

Issue #220, opened 2019-03-11
Run queries in a safe manner


parthshahva:
Currently, we are loading queries into a private s3 bucket and storing a reference in sqs. Things to do:
- [ ] Setup prepared statements or investigate best practice for templating queries
- [ ] Driver should store params and which prepared statement to use in s3
- [x] Use read only user to execute queries from requests

parthshahva:
cc @mweiden as this ticket is based on your suggestions. Free free to update or leave any additional comments

mweiden:
@parthshahva looks good to me!

parthshahva:
5/7 Update: Sticking with readonly user to ensure no non-select queries are run.

According to redshift documentation:
Prepared statements only last for the duration of the current session. When the session ends, the prepared statement is discarded, so it must be re-created before being used again. This also means that a single prepared statement can't be used by multiple simultaneous database clients; however, each client can create its own prepared statement to use. The prepared statement can be manually removed using the DEALLOCATE command.

PR #283, opened 2019-05-07
read only method of running queries

parthshahva:
read-only user setup for query runner related to #220


scripts/redshift/setup_readonly_user.py
```diff
@@ -0,0 +1,25 @@
...
+matrix_config = MatrixRedshiftConfig()
+redshift_handler = RedshiftHandler()
+
+
+def handler():
+    readonly_username = matrix_config.readonly_username
+    readonly_password = matrix_config.readonly_password
+    drop_user_query = f"DROP USER IF EXISTS {readonly_username};"
+    add_user_query = f"CREATE USER {readonly_username} WITH PASSWORD '{readonly_password}';"
+    grant_query = f"grant select on all tables in schema public to {readonly_username};"
```
mckinsel: I think it needs select on `information_schema` to be able to return available fields dynamically.

mckinsel: Well just `information_schema.tables` and `information_schema.columns`

parthshahva: okay will grant select on all tables in information_schema

parthshahva: will "grant select on all tables in schema information_schema" suffice?

mckinsel: Yeah I think that should do it.

scripts/redshift/setup_readonly_user.py
```diff
@@ -0,0 +1,25 @@
...
+    redshift_handler.transaction([drop_user_query])
+    redshift_handler.transaction([add_user_query])
+    redshift_handler.transaction([grant_query])
```
calvinnhieu: Single transaction?

matrix/common/aws/redshift_handler.py
```diff
@@ -44,9 +44,16 @@ def __init__(self):
...
-    def transaction(self, queries: typing.List[str], return_results=False):
+    @property
+    def readonly_database_uri(self):
+        return self.redshift_config.readonly_database_uri
+
+    def transaction(self, queries: typing.List[str], return_results=False, read_only=False):
+        if read_only:
+            conn = pg.connect(self.readonly_database_uri)
+        else:
+            conn = pg.connect(self.database_uri)
```
calvinnhieu: Consider adding a unit test for these cases

PR #284, opened 2019-05-07
level 2 healthcheck checks redshift cluster and ecs query runner

parthshahva:
level 2 healthcheck checks redshift cluster and ecs query runner

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -4,9 +4,14 @@ import os
 import re
 import sys
 
+import boto3
 import chalice
 import connexion
 
+from matrix.common.aws.redshift_handler import RedshiftHandlers
+
+ecs_client = boto3.client("ecs", region_name=os.environ['AWS_DEFAULT_REGION'])
+redshift_handler = RedshiftHandlers()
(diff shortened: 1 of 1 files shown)


chalice/app.py
```diff
@@ -72,6 +77,16 @@ def version():
...
+        # Level 2 healthcheck: Test connection can be made to redshift cluster but do not run any queries
+        redshift_handler.transaction([])
+        # Level 2 healthcheck checks that ecs query runner is active with expected number of tasks
+        service_name = f"matrix-service-query-runner-{os.environ['DEPLOYMENT_STAGE']}"
+        service = ecs_client.describe_services(cluster=service_name, services=[service_name])["services"][0]
+        status = service["status"]
+        desired_task_count = service["desiredCount"]
+        running_task_count = service["runningCount"]
+        assert status == 'ACTIVE'
+        assert desired_task_count == running_task_count
```
calvinnhieu: On fail, will this simply return a 500 to the client?

calvinnhieu:
Looks like the existing unit tests will send network requests. Ideally we'd mock these out to minimize unit test run times.

chalice/app.py
parthshahva: Yea

PR #285, opened 2019-05-07
New API implementation

mckinsel:
This PR implements the v1 API alongside the v0 API. The v1 API is the "metadata and numerical filtering" API. This API allows users to specify some fact about cells via a `filter` property in the request body to `/matrix`. So, they can ask for something like "cells with more than 5000 detected genes" or "cells from the pancreas" rather than having to provide a list of DSS bundle fqids.


matrix/common/constants.py
```diff
@@ -155,3 +162,99 @@ class BundleType(Enum):
...
+    'organ_part_ontology': 'specimen',
+    'organ_part_label': 'specimen',
+    'librarykey': 'library_preparation',
+    'input_nucleic_acid_ontology': 'library_preparation',
+    'input_nucleic_acid_label': 'library_preparation',
+    'construction_approach_ontology': 'library_preparation',
+    'construction_approach_label': 'library_preparation',
+    'end_bias': 'library_preparation',
+    'strand': 'library_preparation',
+    'projectkey': 'project.',
```
calvinnhieu: `project`

matrix/common/query_constructor.py
```diff
@@ -0,0 +1,228 @@
...
+    """Error indicating something wrong with a matrix filter."""
+    pass
+
+
+class MalformedMatrixFeature(Exception):
+    """Error indicating something wrong with a matrix feature."""
+    pass
+
+
+def translate_filters(filter_):
```
calvinnhieu: Could you add type hints?

matrix/common/query_constructor.py
```diff
@@ -0,0 +1,228 @@
...
+    The metadata names that the matrix service presents to users are taken from
+    the project metadata tsvs. These names are different from the table and column
+    names inside the matrix service. This function takes a filter with a field
+    like "project.project_core.project_short_name" and translates it to
+    "project.short_name" so the query will execute correctly.
+    """
+
+    new_filter = dict(filter_)
+    if "field" in filter_:
+        try:
```
calvinnhieu: Can consolidate 107-112 with 132-137

matrix/common/query_constructor.py
```diff
@@ -0,0 +1,228 @@
...
+    "=", ">=", "<=", "<", ">", "!=", "in"
+]
+
+
+LOGICAL_OPERATORS = ["and", "or", "not"]
+
+DEFAULT_FIELDS = ["cell.cell_suspension_id", "cell.genes_detected", "specimen.*",
+                  "library_preparation.*", "project.*", "analysis.*"]
+
+DEFAULT_FEATURE = "gene"
```
calvinnhieu: Use enum

matrix/common/query_constructor.py
```diff
@@ -0,0 +1,228 @@
...
+        "expression_query": expression_query,
+        "cell_query": cell_query,
+        "feature_query": feature_query
+    }
+
+
+def feature_to_where(matrix_feature):
+    """Build the WHERE clause for the features."""
+
+    if matrix_feature == "gene":
```
calvinnhieu: Use enum

matrix/common/query_constructor.py
```diff
@@ -0,0 +1,228 @@
...
+
+        try:
+            field = matrix_filter["field"]
+        except Exception as exc:
+            raise MalformedMatrixFilter("Could not retrieve filter field") from exc
+
+        if op == 'in':
+            if not isinstance(value, (list, tuple)):
+                raise MalformedMatrixFilter("The 'in' operator requires an array value")
+            value = ', '.join([f"'{el}'" if isinstance(el, str) else str(el) for el in value])
```
calvinnhieu: Does `str(el)` need single quotes as well?

matrix/common/query_constructor.py
```diff
@@ -0,0 +1,228 @@
...
+
+def filter_to_where(matrix_filter):
+    """Build a WHERE clause for the matrix request SQL query out of the matrix
+    filter object.
+    """
+
+    try:
+        op = matrix_filter["op"]
+    except Exception as exc:
+        raise MalformedMatrixFilter("Could not retrieve filter op") from exc
```
calvinnhieu: For errors that fail the request, log an error in Dynamo via `request_tracker.log_error` to propagate the failure to the user on GET. These errors could be captured in the driver.

matrix/common/constants.py
```diff
@@ -155,3 +162,99 @@ class BundleType(Enum):
         );
     """
 }
+
+# Map from internal matrix service column names to the names used in the
+# project tsv and hence the API surface.
+TABLE_COLUMN_TO_METADATA_FIELD = {
```
calvinnhieu: Is this intended to have a key for every column in Redshift (to define `FIELD_DETAIL`/`FILTER_DETAIL`? If so, it's missing some for example from `publication` and `contributor` tables.

calvinnhieu: Nevermind, I see the queries in `query_constructor` don't include these tables.

matrix/lambdas/api/v1/core.py
```diff
@@ -0,0 +1,266 @@
...
+
+def get_fields():
+    """Return the list of accepted fields, like
+        ["project.short_name", "library_preparation.strand", ...
+    """
+    return (list(constants.FIELD_DETAIL.keys()),
+            requests.codes.ok)
+
+
+def _redshift_detail_lookup(name, description):
```
calvinnhieu: In order to reduce the chance of API gateway timeouts, could `FIELD_DETAIL_..._QUERY_TEMPLATE`s accept `table_name` and use a single LEFT OUTER JOIN (`PRIMARY_KEY` mapping exists in `RedshiftHandler`)?

tests/functional/test_matrix_service.py
```diff
@@ -179,72 +255,43 @@ def _post_notification(self, bundle_fqid, event_type):
...
-
-        method = getattr(requests, verb.lower())
-        response = method(url, **options)
+        data = json.loads(response)
+        return data["request_id"]
 
-        print(f"-> {response.status_code}")
-        if expected_status:
-            self.assertEqual(expected_status, response.status_code)
+    def test_mtx_output_matrix_service(self):
```
calvinnhieu: To be clear, all previous functional tests are preserved on v0 and v1 is tested with 1 mtx request?

calvinnhieu: It would be valuable to have this test also test `fields` and potentially other `op`s

tests/unit/common/test_query_constructor.py
```diff
@@ -0,0 +1,453 @@
...
+                        "op": "=",
+                        "field": "quuz",
+                        "value": "thud"
+                    }
+                ]
+            }
+        expected_sql = "((foo IN ('bar', 'baz')) OR (qux > 5) OR (quuz = 'thud'))"
+        self.assertEqual(query_constructor.filter_to_where(filter_), expected_sql)
+
+    def test_nested(self):
```
calvinnhieu: 💯

matrix/common/constants.py
mckinsel: Fixed, good catch

matrix/common/query_constructor.py
mckinsel: Yep

matrix/common/query_constructor.py
mckinsel: Fixed

matrix/common/query_constructor.py
mckinsel: Fixed

matrix/common/query_constructor.py
mckinsel: No, you want it to be like `foo IN ("abc", "xyz")` but then `foo IN (1, 2, 3)`

matrix/common/query_constructor.py
mckinsel: Okay, factored it out.

matrix/common/query_constructor.py
mckinsel: Okay, added an `log_error` where this is called in the driver.

matrix/lambdas/api/v1/core.py
mckinsel: Yeah I think that should work.

tests/functional/test_matrix_service.py
mckinsel: Added a couple other functional tests that do this.

mckinsel:
Okay I think first round of comments have been addressed. Ready for another look @calvinnhieu

Issue #236, opened 2019-03-21
Separate out query results and conversion results buckets


parthshahva:
- [x] Separate out query results and conversion results buckets
- [x] Put expiration policy on query results and remove expiration on conversion results

PR #294, opened 2019-05-20
put all query results in separate bucket

parthshahva:
Instead of re-using matrix service results bucket, all query results will now be deposited into their own bucket. This is for ticket #236


terraform/modules/matrix-service/lambdas/driver_lambda.tf
```diff
@@ -120,8 +120,8 @@ resource "aws_lambda_function" "matrix_service_driver_lambda" {
     variables = {
         DEPLOYMENT_STAGE = "${var.deployment_stage}"
         DYNAMO_REQUEST_TABLE_NAME="dcp-matrix-service-request-table-${var.deployment_stage}"
-        MATRIX_QUERY_BUCKET = "dcp-matrix-service-queries-${var.deployment_stage}"
-        MATRIX_RESULTS_BUCKET = "dcp-matrix-service-results-${var.deployment_stage}"
+        MATRIX_QUERY_BUCKET = "dcp-matrix-service-query-results-${var.deployment_stage}"
```
calvinnhieu: In `environment`, this var points to `dcp-matrix-service-queries-*`

terraform/modules/matrix-service/lambdas/driver_lambda.tf
```diff
@@ -120,8 +120,8 @@ resource "aws_lambda_function" "matrix_service_driver_lambda" {
...
+        MATRIX_QUERY_RESULTS_BUCKET = "dcp-matrix-service-results-${var.deployment_stage}"
```
calvinnhieu: I think you wanted to update the value of this to the new bucket

parthshahva: Ah good catch! I made a mess of these variables... will clean up

PR #295, opened 2019-05-22
add ETL support for optimus bundles

calvinnhieu:
Revived `dcp_zarr_store.py` to workaround `!` file separator convention and access DCP zarr structures.

Testing:
- Transformed `d65ef9e6-3fd1-4026-9843-ff573c9e66c8` integration bundle (included bundle in repo for unit testing)
  - Review by commit to reduce Files changed


matrix/common/etl/transformers/cell_expression.py
```diff
@@ -165,6 +172,75 @@ def _parse_10x_bundle(self, bundle_dir):
...
+        Parses optimus analysis files into PSV rows for cell and expression Redshift tables.
+        """
+        keys = self._parse_keys(bundle_dir)
+        cell_suspension_id = json.load(open(
+            os.path.join(bundle_dir, "cell_suspension_0.json")))['provenance']['document_id']
+
+        # read zarr expression matrix
+        store = DCPZarrStore(bundle_dir=bundle_dir)
+        root = zarr.group(store=store)
+        expr_values = root.expression_matrix.expression[:]
```
mckinsel: I think this load the whole dense count array into memory. Are you sure that's necessary? I think it could get pretty big for some bundles.

calvinnhieu: Oh good call! Updated to process zarr in chunks.

matrix/common/etl/dcp_zarr_store.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+"""Class for working with expression matrices serialized in the DCP."""
+
+import glob
+import json
+import os
+from collections import MutableMapping
```
mckinsel: I get a `DeprecationWarning` on this in 3.7, saying it should be `from collections.abc ...`

matrix/common/etl/dcp_zarr_store.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        :param bundle_dir: path to local dir containing analysis bundle
+        """
+        self.bundle_dir = bundle_dir
+        self.cell_suspension_id = json.load(open(
+            os.path.join(self.bundle_dir, "cell_suspension_0.json")))['provenance']['document_id']
+        self.cache = {}
+
+    def __setitem__(self, key, value):
+        raise NotImplementedError("The HCA Data Storage System is read-only.")
```
mckinsel: Just so I understand, this is not pointing at the DSS at all right? It's some local directory that's a result of like `hca download`?

matrix/common/etl/dcp_zarr_store.py
```diff
@@ -0,0 +1,83 @@
...
+    def __delitem__(self, key):
+        raise NotImplementedError("The HCA Data Storage System is read-only.")
+
+    def __getitem__(self, key):
+        if key in self.cache:
+            return self.cache[key]
+
+        transformed_key = self._transform_key(key)
+        path_to_file = os.path.join(self.bundle_dir,
+                                    DCPZarrStore._separator_char.join([f"{self.cell_suspension_id}.zarr",
```
mckinsel: How sure are we that this is the real naming scheme? I think pipeline are kind of free to name these things how they want, creating their own "sample name" for example.

matrix/common/etl/transformers/cell_expression.py
```diff
@@ -37,13 +41,18 @@ def _write_rows_to_psvs(self, *args: typing.Tuple):
 
     def _parse_from_metadatas(self, bundle_dir):
         if os.path.isfile(os.path.join(bundle_dir, "matrix.mtx")):
-            cell_lines, expression_lines = self._parse_10x_bundle(bundle_dir)
+            cell_lines, expression_lines = self._parse_cellranger_bundle(bundle_dir)
+        elif os.path.isfile(os.path.join(bundle_dir, "empty_drops_result.csv")):
```
mckinsel: This feels like a risky way to get the analysis type now that we're starting to get more of them. I think a better way would be like `json.load(open("analysis_protocol_0.json"))["computational_method"] == "Optimus"`, but I'm checking with the pipelines team.

matrix/common/etl/transformers/cell_expression.py
```diff
@@ -165,9 +174,108 @@ def _parse_10x_bundle(self, bundle_dir):
...
+        """
+        keys = self._parse_keys(bundle_dir)
+
+        # read expression matrix from zarr
+        store = DCPZarrStore(bundle_dir=bundle_dir)
+        root = zarr.group(store=store)
+
+        n_cells = root.expression_matrix.cell_id.shape[0] if not test_n_cells else test_n_cells
+        chunk_size = root.expression_matrix.expression.chunks[0] if not test_n_cells else test_n_cells - 1
+        n_chunks = int(math.ceil(n_cells / chunk_size))
```
mckinsel: I think this is just `root.expression_matrix.expression.nchunks`

matrix/common/etl/dcp_zarr_store.py
calvinnhieu: Good catch

matrix/common/etl/dcp_zarr_store.py
calvinnhieu: True (technically GET file). Updating message to "DCPZarrStore is read-only" since we don't have a use for writing currently.

matrix/common/etl/dcp_zarr_store.py
calvinnhieu: I'm not sure, but that's fair - updated to read file paths instead of assuming

matrix/common/etl/transformers/cell_expression.py
calvinnhieu: Sounds good. Will update when we have an answer

matrix/common/etl/transformers/cell_expression.py
```diff
@@ -165,9 +174,108 @@ def _parse_10x_bundle(self, bundle_dir):
...
+                 keys["specimen_key"],
+                 keys["library_key"],
+                 keys["analysis_key"],
+                 barcodes[i],
+                 str(gene_count)]
+            ) + '\n'
+            cell_lines.add(cell_line)
+
+            # skip 0 counts
+            if gene_count == 0:
```
mckinsel: Does this happen? It would be a weird thing to see in the optimus result I think.

calvinnhieu: Oh doesn't look like it does. Removing

matrix/common/etl/transformers/cell_expression.py
mckinsel: Okay, so based on their response I would do this:
```python
protocol_id = json.load(open(os.path.join(bundle_dir, "analysis_protocol_0.json")))["protocol_core"]["protocol_id"]
if protocol_id.startswith("optimus"):
    ....
elif protocol_id.startswith("cellranger"):
    ....
elif protocol_id.startswith("smartseq2"):
    ....
````

File: tests/unit/common/etl/transformers/test_cell_expression.py [hunk 8]
```diff
@@ -23,3 +24,17 @@ def test_parse_from_metadatas(self):
         expression_rows = parsed[1][1]
         self.assertEqual(expression_table, TableName.EXPRESSION)
         self.assertEqual(expression_rows[0], "635badd5-7d62-4db3-b509-f290a12a1336|ENST00000373020|TPM|92.29\n")
+
+    def test_parse_optimus(self):
+        cell_lines, expression_lines = self.transformer._parse_optimus_bundle(
+            bundle_dir=os.path.abspath("tests/functional/res/etl/d65ef9e6-3fd1-4026-9843-ff573c9e66c8"),
+            test_n_cells=6
```
mckinsel: Can you use `mock` here instead of including a test parameter?

calvinnhieu: Modified test data and removed test parameter

matrix/common/etl/transformers/cell_expression.py
calvinnhieu: Done

tests/functional/res/etl/d65ef9e6-3fd1-4026-9843-ff573c9e66c8/493a6adc-54b5-4388-ba11-c37686562127.zarr!expression_matrix!expression!.zarray
```diff
@@ -1,7 +1,7 @@
 {
     "chunks": [
-        10000,
-        10000
+        5,
+        20000000
```
mckinsel: What does this do? It seems like this should make the tests fail?

calvinnhieu: Oh yes this is weird. Updated to only restrict `n_cells`

calvinnhieu: Oh yes this does produce incorrect output. Returned to normal chunk sizes

calvinnhieu:
Yup that makes sense

PR #296, opened 2019-05-24
handle DSS update events

### LINKED ISSUE #282: Update notifications to handle DSS UPDATE events
Per the requirement in the [Simple Metadata Updates RFC](https://github.com/HumanCellAtlas/dcp-community/blob/0d35fdc8011a84a7f2de0fe8e2e55b776716f1a5/rfcs/text/0000-simple-updates.md#dcp-flow-of-updates):

>Once the bundles are updated (1 & 2), the datastore will send update notifications to all listeners (3). The query service, data browser and matrix service MUST all process these notifications to provide the updated metadata to consumers as required. 

DoD:
- [ ] UPDATE events trigger a re-ETL of a given bundle UUID and uses the new bundle version

calvinnhieu:
The ETL library queries on `bundle_uuid` and downloads the latest version. The notification handler will replace the old bundle in Redshift with the new bundle.

Closes #282

### DIFF
--- a/matrix/lambdas/daemons/notification.py
+++ b/matrix/lambdas/daemons/notification.py
@@ -21,7 +21,7 @@ class NotificationHandler:
         self.redshift = RedshiftHandler()
 
     def run(self):
-        if self.event_type == 'CREATE':
+        if self.event_type == 'CREATE' or self.event_type == 'UPDATE':
             self.update_bundle()
         elif self.event_type == 'DELETE' or self.event_type == 'TOMBSTONE':
             self.remove_bundle()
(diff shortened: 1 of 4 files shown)


mckinsel:
Are there any ways to get slightly more thorough functional tests for the notifications? Like maybe DELETE, make an API request, and verify that you only get four bundles. Then UPDATE and show that now you get five?

calvinnhieu:
I agree that the validation for the presence/absence of a bundle could be improved (i.e. could verify more tables such as cell and expression row counts), but I would prefer to not run more API requests to keep functional test times to a minimum.

Addressing the issue we saw with missing bundles due to interrupting the notification test, I think submitting a `CREATE` for the test bundle used in the notification test in `tearDownClass()` would resolve this. What do you think @mckinsel?

calvinnhieu:
Updated to reflect above proposed changes

PR #297, opened 2019-05-29
Fix detail requests for fields in the cell table

mckinsel:
Previously, requests for details about a field in the cell table, like `genes_detected` for example, would generate sql with `...FROM cell LEFT OUTER JOIN cell...`. This fixes that.

### DIFF
--- a/tests/unit/common/test_query_constructor.py
+++ b/tests/unit/common/test_query_constructor.py
@@ -451,3 +451,52 @@ MANIFEST VERBOSE
 ;
 """)
         self.assertEqual(queries["expression_query"], expected_exp_query)
+
+
+class TestDetailQuery(unittest.TestCase):
+
+    def test_cell_numeric(self):
+        query = query_constructor.create_field_detail_query(
+            "cell.genes_detected",
+            "cell",
+            "cellkey",
+            "numeric")
+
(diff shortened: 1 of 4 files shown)


tests/unit/common/test_query_constructor.py
```diff
@@ -451,3 +451,52 @@ def test_filter_conversion(self):
...
+            "cell",
+            "cellkey",
+            "categorical")
+
+        expected_sql = """
+SELECT cell.cell_suspension_id, COUNT(cell.cellkey)
+FROM cell 
+GROUP BY cell.cell_suspension_id
+;
+"""  # noqa: W291
```
calvinnhieu: nit: can't you just remove the whitespace?

https://lintlyci.github.io/Flake8Rules/rules/W291.html

mckinsel: The template has a `FROM cell {join}\n`, so if `join` is empty, there's a space at the end of the line. I could make it `FROM cell{join}\n` and put a whitespace in the beginning of the string passed to `join`, but I thought that looked weirder.

calvinnhieu: Oh nevermind me then. Thought the violation was on line 485

Issue #221, opened 2019-03-11
Investigate how to ensure bundles are not silently dropped from output


parthshahva:
Questions:
1) What is the best way to maintain an "expected_cells" value for a request?
2) Should we return an error instead of serving matrix if cells do not match expected_cells count?
3) Should we ever be okay with dropping bundles from the output?

mckinsel:
I think this is going to be pretty difficult, since we won't know what we don't know. But one thing that we could check is that the count of distinct cellkeys selected from the expression table matches the count of rows selected from the cell table.

parthshahva:
Solution: compare count(*) on analysis to number of input bundles. Return error if there is a mismatch

parthshahva:
@calvinnhieu would this query consistently return within ~15 seconds with short query acceleration in place?

parthshahva:
Actually nevermind, this query is going to run in the driver so it has a lot longer (will increase driver lambda timeout to 15 minutes)

PR #306, opened 2019-06-04
check request bundles exist in matrix service db

parthshahva:
This PR accomplishes #221. We will now return a failure in v0 driver if the request bundles do not all exist within our database (specifically the analysis table)

### DIFF
--- a/matrix/lambdas/daemons/v0/driver.py
+++ b/matrix/lambdas/daemons/v0/driver.py
@@ -7,10 +7,17 @@ from matrix.common.config import MatrixInfraConfig, MatrixRedshiftConfig
 from matrix.common.logging import Logging
 from matrix.common.request.request_tracker import RequestTracker, Subtask
 from matrix.common.aws.dynamo_handler import DynamoHandler, DynamoTable, RequestTableField
+from matrix.common.aws.redshift_handler import RedshiftHandler
 from matrix.common.aws.sqs_handler import SQSHandler
 from matrix.common.aws.s3_handler import S3Handler
 
 logger = Logging.get_logger(__name__)
+
+analysis_bundle_count_query_template = """
+    $$SELECT count(*)
+    FROM analysis
+    WHERE bundle_fqid IN {0}$$
+"""
 
 expression_query_template = """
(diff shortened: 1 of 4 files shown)


mckinsel:
Do you think we should do anything in the v1 api too? There's an equivalent situation with a query like `{"op": "in", "field": "dss_bundle_fqid", "value": ["fake-bundle-1", "fake-bundle-2"]}`

matrix/lambdas/daemons/v0/driver.py
```diff
@@ -104,6 +112,14 @@ def run(self, bundle_fqids: typing.List[str], bundle_fqids_url: str, format: str
                                                        RequestTableField.NUM_BUNDLES,
                                                        len(resolved_bundle_fqids))
         s3_obj_keys = self._format_and_store_queries_in_s3(resolved_bundle_fqids)
+
+        analysis_table_bundle_count = self._fetch_bundle_count_from_analysis_table(resolved_bundle_fqids)
+        if analysis_table_bundle_count != len(resolved_bundle_fqids):
+            error_msg = "resolved bundles in request do not match bundles available in matrix service"
+            logger.info(error_msg)
```
calvinnhieu: `log_error` also runs `logger.debug`

matrix/lambdas/daemons/v0/driver.py
```diff
@@ -149,3 +165,10 @@ def _add_request_query_to_sqs(self, s3_obj_key: str):
         }
         logger.debug(f"Adding {payload} to sqs {queue_url}")
         self.sqs_handler.add_message_to_queue(queue_url, payload)
+
+    def _fetch_bundle_count_from_analysis_table(self, resolved_bundle_fqids: list):
+        analysis_table_bundle_count_query = analysis_bundle_count_query_template.format(
+            self._format_bundle_fqids(resolved_bundle_fqids))
```
calvinnhieu: FYI I refactored this out of the Driver here: https://github.com/HumanCellAtlas/matrix-service/pull/308/commits/6f103c46a198b9935483bec7fefed7a32af40060

calvinnhieu:
Looks good to me. +1 to Marcus' suggestion - I think it makes sense to deliver the same experience for anyone migrating from v0 to v1

PR #308, opened 2019-06-04
Update DELETE notification handler to support v1 API

calvinnhieu:
Changes:
- DELETE now deletes all rows in `cell`, `expression` and `analysis` tables associated with the deleted bundle (previously only `analysis`, i.e. the reference to the bundle fqid)
- Updated functional test to verify row counts by cellkey (previously bundle fqid)
- Refactored `Driver._format_bundle_fqids` -> `query_constructor.list_to_query_str`

Manual testing:
- functional tests passing locally

### DIFF
--- a/matrix/common/query_constructor.py
+++ b/matrix/common/query_constructor.py
@@ -250,3 +250,7 @@ def filter_to_where(matrix_filter: typing.Dict[str, typing.Any]) -> str:
                 f"({filter_to_where(v)})" for v in value]) + ')'
     else:
         raise MalformedMatrixFilter(f"Invalid op: {op}")
+
+
+def list_to_query_str(values: typing.Iterable[str]) -> str:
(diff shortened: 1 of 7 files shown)


matrix/common/query_constructor.py
```diff
@@ -250,3 +250,7 @@ def filter_to_where(matrix_filter: typing.Dict[str, typing.Any]) -> str:
                 f"({filter_to_where(v)})" for v in value]) + ')'
     else:
         raise MalformedMatrixFilter(f"Invalid op: {op}")
+
+
+def list_to_query_str(values: typing.Iterable[str]) -> str:
```
mckinsel: This will return an incorrect string if `values` is numeric. Do you think it's worth flagging or checking that?

calvinnhieu: The type hint specifies `str` as input, but I agree it could be clearer. Renamed to `format_str_list` + added comment

PR #311, opened 2019-06-07
Add support for per project ETL

### LINKED ISSUE #304: Update loader tool to support per project updates
DoD:
- [x] Loader can incrementally load individual or multiple projects in successive executions (currently the loader can only be run once and assumes the database is empty)
- [x] Loader script exposes an input parameter to specify a list of project uuids to load in (if no project uuids are supplied, perform a full load)

calvinnhieu:
- provides `--project-uuids` input parameter to specify project UUIDs to ETL
- row verification after load
- port loader script from bash to python

Closes #304 

Testing:
- Loaded `dev` and `prod` environments 1 project at a time


scripts/redshift/loader.py
```diff
@@ -0,0 +1,155 @@
...
+
+from matrix.common.aws.redshift_handler import RedshiftHandler
+from matrix.common.query_constructor import format_str_list
+from matrix.common.etl import (get_dss_client,
+                               etl_dss_bundles,
+                               transform_bundle,
+                               finalizer_reload,
+                               finalizer_update,
+                               upload_and_load,
+                               load_tables)
```
mckinsel: I know this isn't really the style that's used throughout the rest of the project, but at some point it might be easier to just `from matrix.common import etl`

scripts/redshift/loader.py
```diff
@@ -0,0 +1,155 @@
...
+    }
+]
+
+
+def _build_dss_query(project_uuids):
+    q = DSS_SEARCH_QUERY_TEMPLATE
+
+    if project_uuids:
+        for uuid in project_uuids:
+            q['query']['bool']['must'][0]['bool']['should'].append({
```
mckinsel: What does the nesting of `should` and `must` get you?

calvinnhieu: Hm, it's enabling multiple `should` clauses (for which a should clause implements an OR to support multiple key names). The top level `must` is an array, as opposed to just a `bool` clause is a dict.

PR #330, opened 2019-06-25
return error from post request and driver of v0 if bundle manifest or bundle input is empty

### LINKED ISSUE #327: Handle empty v0 requests
You can submit v0 requests with an empty bundle list, and it will run for a while and then fail. The post request should 400 right away instead.

parthshahva:
Resolves #327 . Return error from post request and driver of v0 if bundle manifest or bundle input is empty

### DIFF
--- a/matrix/lambdas/api/v0/core.py
+++ b/matrix/lambdas/api/v0/core.py
@@ -26,24 +26,24 @@ def post_matrix(body: dict):
     if format not in expected_formats:
         return ({'message': "Invalid parameters supplied. "
                             "Please supply a valid `format`. "
-                            "Visit https://matrix.dev.data.humancellatlas.org for more information."},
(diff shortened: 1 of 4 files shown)


matrix/lambdas/api/v0/core.py
```diff
@@ -52,6 +52,11 @@ def post_matrix(body: dict):
     else:
         bundle_fqids = body['bundle_fqids']
         bundle_fqids_url = None
+        if len(bundle_fqids) == 0:
+            return ({'message': "Invalid parameters supplied. "
+                                "Please supply non empty `bundle_fqids`. "
+                                "Visit https://matrix.dev.data.humancellatlas.org for more information."},
```
mckinsel: Shouldn't need the `.dev.` right?

parthshahva: fixed

Issue #231, opened 2019-03-15
Write vignette for Seurat


brianraymor:
***Goal**: Publish code samples/vignettes for loading matrix service data into three different single cell analysis toolkits. This will be modeled on the scanpy vignette.*

PR #333, opened 2019-07-02
matrix service to seurat vignette

parthshahva:
Vignette for #231


calvinnhieu:
Aside from comments on Slack (investigate nbviewer inconsistencies, operations on Seurat object https://satijalab.org/seurat/v3.0/pbmc3k_tutorial.html), you could also reset and rerun the Notebook to annotate the steps starting from 1. Also note spacing in step [148]. Other than that, looks good to me!

mckinsel:
Looks good, two small things:
- Can you install seurat via `install.packages('Seurat')`? I think that might do less compilation.
- Remove the line that opens the readme; it's not in the zip anymore.

PR #354, opened 2019-07-25
Add UMI count filter to optimus ETL

mckinsel:
The optimus zarr output contains many `cell_id`s that are unlikely to be genuine cells. For example, in one example bundle, there are ~230,000 elements in `cell_id`, but emptyDrops suggests that about 3,000 of those are real cells.

The matrix service should strike a balance between allowing users to download and then perform their own filtering on the full set of barcodes and not delivering lots of low-quality data. This PR implements a permissive filter on data loaded into the matrix service from optimus bundles. It makes the matrix service ignore `cell_id`s with a UMI count less than 100, as these are almost certainly artifactual.

This PR also adds two new fields to the matrix service metadata: `total_umis` and `emptrydrops_is_cell`. The 100 minimum count filter is probably less restrictive than most users will want, and these two new fields will help users who want to perform subsequent filtering.


matrix/common/etl/transformers/cell_expression.py
```diff
@@ -239,7 +256,9 @@ def _parse_optimus_chunk(self,
                  keys["library_key"],
                  keys["analysis_key"],
                  barcodes[i],
-                 str(gene_count)]
+                 str(gene_count),
+                 str(emptydrops_result[barcodes[i]]["total_umi_count"]),
+                 't' if emptydrops_result[barcodes[i]]["is_cell"] else 'f']
```
calvinnhieu: `emptydrops_is_cell` in table schema is `BOOLEAN`

matrix/common/constants.py
```diff
@@ -52,6 +52,8 @@ class BundleType(Enum):
             analysiskey        VARCHAR(60) NOT NULL,
             barcode            VARCHAR(32),
             genes_detected     INTEGER,
+            total_umis         INTEGER,
+            emptydrops_is_cell BOOLEAN,
```
calvinnhieu: Can these be more explicit about being tied to Optimus? I considered a comment here (I usually use this file as a quick reference to the current table schema), or even better if the column name could reflect Optimus?

matrix/common/etl/transformers/cell_expression.py
```diff
@@ -16,6 +16,8 @@
 from matrix.common.etl.dcp_zarr_store import DCPZarrStore
 from matrix.common.exceptions import MatrixException
 
+EMPTYDROPS_MIN_COUNT = 100
```
calvinnhieu: Can we make this a class variable in the class?

matrix/common/etl/transformers/cell_expression.py
```diff
@@ -96,7 +98,9 @@ def _parse_ss2_bundle(self, bundle_dir):
             keys["library_key"],
             keys["analysis_key"],
             "",
-            str(genes_detected)]) + '\n']
+            str(genes_detected),
+            "",
+            ""]) + '\n']
```
calvinnhieu: `emptydrops_is_cell` in table schema is `BOOLEAN`

matrix/common/etl/transformers/cell_expression.py
```diff
@@ -174,7 +178,9 @@ def _parse_cellranger_bundle(self, bundle_dir):
                  keys["library_key"],
                  keys["analysis_key"],
                  cell_to_barcode[cell_key],
-                 str(gene_count)]) + '\n'
+                 str(gene_count),
+                 "",
+                 ""]) + '\n'
```
calvinnhieu: `emptydrops_is_cell` in table schema is `BOOLEAN`

tests/unit/common/etl/transformers/test_cell_expression.py
```diff
@@ -57,10 +57,11 @@ def test_parse_optimus(self):
             bundle_dir=os.path.abspath("tests/functional/res/etl/optimus_bundle")
         )
         self.assertEqual(len(cell_lines), 5)
+        print(cell_lines)
```
calvinnhieu: print

matrix/common/etl/transformers/cell_expression.py
mckinsel: Yeah, the redshift docs say 't' or 'f' are okay literal values for boolean fields. https://docs.aws.amazon.com/redshift/latest/dg/r_Boolean_type.html (I think I'm understanding that right)

matrix/common/constants.py
mckinsel: Yeah, actually would be interested in a suggestion from the pipelines team on this one. I think these fields should apply to all droplet scRNA-seq though.

matrix/common/etl/transformers/cell_expression.py
mckinsel: Yeah, will fix.

matrix/common/etl/transformers/cell_expression.py
calvinnhieu: Oh I see!

tests/unit/common/etl/transformers/test_cell_expression.py
mckinsel: Ah forgot to delete that.

matrix/common/etl/transformers/cell_expression.py
```diff
@@ -20,6 +20,10 @@
 class CellExpressionTransformer(MetadataToPsvTransformer):
     """Reads SS2 and 10X bundles and writes out rows for expression and cell tables in PSV format."""
 
+    # The minimum UMI count in the emptydrops result required to include a
+    # putative cell in the matrix service.
+    emptydrops_min_count = 100
```
nh3: Would be great to allow the users to adjust this parameter from the API!

mckinsel: Hey @nh3, so the user can adjust this via the API through the `total_umis` filter. They just couldn't set that value lower than 100. Thus far, most of the feedback that we've gotten is that 100 UMIs is a safe lower bound for users who want to query the matrix service for _cells_, which is the main use case we're trying to solve. Users who want the full set of droplets usually want to go back to raw data anyway, so wouldn't really be using matrix service outputs.

PR #356, opened 2019-07-29
Add bundle/file uuid/version to matrix metadata

### LINKED ISSUE #291: Add bundle/file uuids/versions to matrix metadata output
To track the provenance of a matrix service output file we need to add 4 additional fields: bundle uuid, bundle version, file uuid, file version.
This would allow the user to correlate any matrix service output file back to the corresponding manifest.
The cell_suspension_id is not versioned and is therefore not sufficient.

mckinsel:
This allows the metadata to be joined with the project-level metadata TSV files.

Resolves #291 .


matrix/common/etl/transformers/cell_expression.py
```diff
@@ -88,13 +88,20 @@ def _parse_ss2_bundle(self, bundle_dir):
 
         genes_detected = sum((1 for k in genes_values.values() if k["Count"] > 0))
 
+        file_uuid = [f for f in json.load(open(bundle_manifest_path))["files"]
+                     if f["name"].endswith(".genes.results")][0]["uuid"]
+        file_version = [f for f in json.load(open(bundle_manifest_path))["files"]
+                        if f["name"].endswith(".genes.results")][0]["version"]
```
calvinnhieu: Can we have/is the choice of `.genes.results`, `matrix.mtx` and `.zattrs` as the identifying file documented somewhere more public than the code here for developer purposes?

mckinsel: No not currently, I'll write something up we can put on the repo and maybe on the browser.

PR #367, opened 2019-08-12
pin supported metadata schema versions in Redshift loader

### LINKED ISSUE #341: Update subscription to DSS to only receive data adhering to compatible schemas
This is required in order to receive data that adheres to a schema versions with which matrix service is compatible.

calvinnhieu:
Closes #341 

In accordance with Phase 1.4 of the [Querying DSS by Metadata Schema Version(s) RFC](https://github.com/HumanCellAtlas/dcp-community/pull/90/files?short_path=41a4263#diff-41a4263df7296b16ac19395c05f3ebca)

### DIFF
--- a/scripts/redshift/loader.py
+++ b/scripts/redshift/loader.py
@@ -20,38 +20,54 @@ class MetadataSchemaName(Enum):
     ORGANOID = "organoid"
 
 
-LATEST_SUPPORTED_MD_SCHEMA_VERSIONS = {
+SUPPORTED_METADATA_SCHEMA_VERSIONS = {
     MetadataSchemaName.PROJECT: {
-        'major': 14,
-        'minor': 0
+        'max_major': 14,
+        'max_minor': 0,
+        'min_major': 1,
+        'min_minor': 0
     },
     MetadataSchemaName.LIBRARY_PREPARATION_PROTOCOL: {
-        'major': 6,
-        'minor': 1
+        'max_major': 6,
+        'max_minor': 1,
+        'min_major': 1,
+        'min_minor': 0
     },
     MetadataSchemaName.ANALYSIS_PROTOCOL: {
-        'major': 9,
-        'minor': 0
+        'max_major': 9,
+        'max_minor': 0,
+        'min_major': 1,
+        'min_minor': 0
     },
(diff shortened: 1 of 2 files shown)


scripts/redshift/loader.py
```diff
@@ -2,11 +2,59 @@
...
+    },
+    MetadataSchemaName.LIBRARY_PREPARATION_PROTOCOL: {
+        'major': 6,
+        'minor': 1
+    },
+    MetadataSchemaName.ANALYSIS_PROTOCOL: {
+        'major': 9,
+        'minor': 0
+    },
+    MetadataSchemaName.SPECIMEN_FROM_ORGANISM: {
```
mckinsel: We also have minimum supported versions. For example, for `specimen_from_organism`, it's 9.0.0.

scripts/redshift/loader.py
```diff
@@ -96,6 +145,57 @@ def load(args):
...
+    :param schema_name: MetadataSchemaName to generate clause for
+    :return: dict ES query clause
+    """
+    return {
+        "bool": {
+            "should": [
+                {
+                    "bool": {
+                        "must_not": {
+                            "exists": {
```
mckinsel: Maybe we can add some comments around these? To me (with not much ES knowledge), this looks like we don't want bundles that have the `schema_major_version` field. Does that mean we would accept bundles without the schema version annotated this way?

calvinnhieu: Sure, will add comments. 

Yes, specifically, this `must_not` `exists` is in an "OR" (`should`) clause that will match bundles that either 1) have `schema_*_version` fields defined and is within our supported version range or 2) do not have `schema_*_version` fields defined. 

The importance of (2) is to continue to match pre-schema-version bundles before existing bundles are backfilled with the new schema fields. This shouldn't be an issue as we currently support these data.

scripts/redshift/loader.py
calvinnhieu: Gotcha. Do you have a list of minimum supported versions for all schemas?

mckinsel: Talked to @maniarathi about this on slack. The answer is that I do not have a list, and it's pretty tedious to create one. But, the minimum for `specimen` is probably the only relevant one.

PR #398, opened 2019-09-27
Add bundle fqid support to loader

calvinnhieu:


mckinsel:
I would have thought either (`manifest.uuid` and `manifest.version`) or (`uuid` and `version`) would be the right terms. Is it possible that they all work?

calvinnhieu:
Hm, I don't get results from `/search` using the fields `manifest.uuid`, but I do using:
```
{
"es_query": {
  "query": {
    "bool": {
      "must": [{
"match": {
                    "manifest.uuid": "ffd3bc7b-8f3b-4f97-aa2a-78f9bac93775"
                }
      }]
    }
  }
}
}
```
Similar for `version` (not working) and `manifest.version` (working)

mckinsel:
Huh.....okay then

PR #408, opened 2019-10-14
Update mtx format to be 10x compatible

### LINKED ISSUE #324: Make a format that is compatible with `read_10x` methods
Lots of tools (seurat, scanpy, etc) have methods to read outputs from cellranger. The matrix service `mtx` output is similar to cellranger but not similar enough to use the existing methods. We should either tweak the mtx output to work or create a `cellranger` or `10x` format.

calvinnhieu:
Modifies/adds files to `mtx` output to support [10x compatibility](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/output/matrices) (e.g. `scanpy.read_10x_mtx`, `seurat.Read10X`).

- ~adds `featuretype_10x` column `feature` table~
- appends `featuretype_10x` column during mtx conversion
- renames `genes.tsv.gz` to `features.tsv.gz` (removes column headers)
- adds `barcodes.tsv` (no column headers)

Resolves #324


matrix/common/constants.py
```diff
@@ -158,6 +158,7 @@ class MetadataSchemaName(Enum):
         CREATE {0}TABLE IF NOT EXISTS {2} (
             featurekey       VARCHAR(25) NOT NULL,
             featurename      VARCHAR(40) NOT NULL,
+            featuretype_10x  VARCHAR(20),
```
calvinnhieu: Requires table reindex

tests/functional/validation.py
```diff
@@ -108,6 +109,7 @@ def calculate_ss2_metrics_mtx(mtx_zip_url):
         shutil.copyfileobj(response.raw, local_mtx_zip_file)
 
     mtx_zip = zipfile.ZipFile(local_mtx_zip_path)
+    scanpy.read_10x_mtx(local_mtx_zip_path)
```
calvinnhieu: TODO: clean up/refactor this functional test

calvinnhieu:
Unable to add functional tests until persistent 10x test bundles are available in dev environments.

matrix/common/constants.py
calvinnhieu: Adding this data on-the-fly, table reindex not required.

tests/functional/validation.py
calvinnhieu: Unable to add functional tests until persistent 10x test bundles are available in dev environments.

calvinnhieu:
Verified 10X compatibility via both `scanpy.read_10x_mtx`  `seurat.Read10X`

matrix/lambdas/api/v1/core.py
```diff
@@ -51,6 +51,10 @@ def post_matrix(body: dict):
         # Otherwise, default to human-only
         genera_species = [constants.GenusSpecies.HUMAN]
 
+    # Cell barcode is required in .mtx matrices by 10x specifications
+    if format_ == MatrixFormat.MTX.value and "cell.barcode" not in fields:
```
mckinsel: I think you want `"cell.barcode" not in fields and "barcode" not in fields`. If it's the latter, it will get translated to `cell.barcode` in the query_constructor.

calvinnhieu: Oh I see. I thought `cell.barcode` was the user-facing field since it's the default name used in both APIs, but `barcode` is used everywhere else (`filters` endpoint, queries). I believe we want to just stop using `cell.barcode` everywhere. Note that most matrices' metadata have been generated with `cell.barcode` and this may break downstream analyses.

calvinnhieu: Nevermind, I see what you're saying re: `query_constructor._get_internal_name`. Going with suggestion

calvinnhieu:
> One thing I want to consider is if the change from genes.tsv.gz to features.tsv.gz is enough of a breaking change that we should preserve the old behavior somewhere.

I agree - reintroduced `genes.tsv.gz`. I initially removed it to avoid confusion, but breaking changes would be worse. Were we ever exporting a README in mtx? Would be good to document usage.

mckinsel:
Yeah we did include a readme in mtx and csv for a while. We took them out because we had to remove it from the loom output. I don't think this needs to be addressed in this PR, but I think the better overall solution would be to make `format` a little more descriptive of the downstream tools you can use it with, so like `10x_v3_mtx` or `seurat_r_obj` or whatever. Just saying `mtx` or `csv` leaves it a little underspecified.

brianraymor:
Agreed. We need something more like [How is the MEX format used for the gene-barcode matrices?](https://kb.10xgenomics.com/hc/en-us/articles/115000794686-How-is-the-MEX-format-used-for-the-gene-barcode-matrices-)

PR #418, opened 2019-10-31
Split out donor and cell suspension concepts into separate tables

mckinsel:
Previously, donor and cell suspension information was pushed into the
specimen table. But this creates problems modeling experiments where,
say, a single specimen is turned into a cell suspension in different
ways.

This change creates two new tables, donor and cell_suspension. It also
adds new tests to the speciment/donor/library/cell_suspension
tranformers.


matrix/common/constants.py
```diff
@@ -49,10 +49,10 @@ class MetadataSchemaName(Enum):
     ORGANOID = "organoid"
 
 
-DEFAULT_FIELDS = ["cell.cell_suspension_id", "cell.genes_detected", "cell.file_uuid",
+DEFAULT_FIELDS = ["cell.cellsuspensionkey", "cell.genes_detected", "cell.file_uuid",
                   "cell.file_version", "cell.total_umis", "cell.emptydrops_is_cell",
-                  "cell.barcode", "specimen.*", "library_preparation.*", "project.*",
-                  "analysis.*"]
+                  "cell.barcode", "cell_suspension.*", "specimen.*", "donor.*",
```
calvinnhieu: Will this include both `cell.cellsuspensionkey` and `cell_suspension.cellsuspensionkey` in the matrix? I suppose it won't cause confusion but unnecessary.

matrix/common/constants.py
```diff
@@ -251,26 +278,33 @@ class MetadataSchemaName(Enum):
...
     'specimenkey': 'specimen_from_organism.provenance.document_id',
-    'genus_species_ontology': 'specimen_from_organism.genus_species.ontology',
-    'genus_species_label': 'specimen_from_organism.genus_species.ontology_label',
+    'donorkey': 'donor_organism.provenance.document_id',
+    'genus_species_ontology': 'cell_suspension.genus_species.ontology',
+    'genus_species_label': 'cell_suspension.genus_species.ontology_label',
     'ethnicity_ontology': 'donor_organism.human_specific.ethnicity.ontology',
     'ethnicity_label': 'donor_organism.human_specific.ethnicity.ontology_label',
     'disease_ontology': 'donor_organism.diseases.ontology',
     'disease_label': 'donor_organism.diseases.ontology_label',
```
calvinnhieu: `diseases`?

matrix/common/constants.py
```diff
@@ -480,10 +521,13 @@ class MetadataSchemaName(Enum):
...
         "Unique identified for the specimen that was collected from the donor organism.",
-    "specimen_from_organism.genus_species.ontology":
-        "An ontology term identifier in the form prefix:accession for the species to which the donor organism belongs.",
-    "specimen_from_organism.genus_species.ontology_label":
-        "The preferred label for the specimen_from_organism.genus_species.ontoloty ontology term",
+    "cell_suspension.genus_species.ontology":
+        ("An ontology term identifier in the form prefix:accession for the species to which the cell(s) "
+         "in the cell suspension belong."),
+    "cell_suspension.genus_species.ontology_label":
+        "The preferred label for the cell_suspension.genus_species.ontology ontology term",
```
calvinnhieu: ontology twice

matrix/common/constants.py
```diff
@@ -182,26 +181,54 @@ class MetadataSchemaName(Enum):
...
             PRIMARY KEY(specimenkey))
             DISTSTYLE ALL
             SORTKEY(specimenkey)
         ;
     """,
+    'cell_suspension': """
+        CREATE {0}TABLE IF NOT EXISTS {2} (
+            cellsuspensionkey              VARCHAR(40) NOT NULL,
+            specimenkey                    VARCHAR(40) NOT NULL,
+            donorkey                       VARCHAR(40) NOT NULL,
```
calvinnhieu: Is there a rationale in linking `cell_suspension` directly to both `specimen` and `donor` as opposed to `cell_suspension` -> `specimen` -> `donor`?

From what I can tell, the former requires one less join to access `donor` information but the latter is more explicit in representing the entity graph. Thoughts?

matrix/common/etl/transformers/specimen_library.py
```diff
@@ -30,49 +42,125 @@ def _write_rows_to_psvs(self, *args: typing.Tuple):
     def _parse_from_metadatas(self, search_dir: str):
```
calvinnhieu: Not crucial for this PR, but we should definitely refactor the metadata transformers (excludes `cell_expression` transformers which technically parse more than metadata) to favor modularity over performance. Transformer groups were initially defined by the metadata files they consumed in order to avoid processing the same file multiple times (e.g. ontology determination in the `specimen` table depends on `library` metadata creating `specimen_library.py`).

Favoring modularity proposes defining a transformer per table. Considering the limited number of rows we can expect per project for `specimen`, `donor`, `library`, `project`, `publication` and `contributor` entities, processing these entities multiple times will not affect performance. I propose writing one transformer for each of these entities, as well as `cell_suspension` (which will be processed once per the table structure in this PR), to improve readability of the code without affecting performance.

matrix/common/etl/transformers/specimen_library.py
```diff
@@ -91,13 +179,18 @@ def resolve_ontology_ols(self, term):
         try:
             ontology, term_id = term.split(":")
         except ValueError:
-            return None
+            try:
+                # Well this isn't right, but it's present in the metadata
+                ontology, term_id = term.split("_")
```
calvinnhieu: i.e this is an upstream data sanitization problem?

matrix/common/constants.py
mckinsel: Yep

matrix/common/constants.py
mckinsel: 👍

matrix/common/constants.py
mckinsel: Yeah I did it because there was one fewer join. That said, I'm not really sure if that gets us anything worthwhile. Can definitely switch to the cs -> specimen -> donor if you think that's clearer.

matrix/common/etl/transformers/specimen_library.py
mckinsel: Yeah I agree, this class is now kind of a random hodgepodge of code. It will probably get worse if we need to start creating more "derived" fields.

matrix/common/etl/transformers/specimen_library.py
mckinsel: Yes it is. It is supposed to be fixed eventually, and in theory new validation tools should be preventing it now.

matrix/common/constants.py
calvinnhieu: Yeah I'd prefer that. Sorry, looks like you'll have to reload the projects then?

matrix/common/etl/transformers/specimen_library.py
calvinnhieu: Any way for us to know when this is fixed in prod data & we can remove?

mckinsel: Apparently there will be a ticket in ingest-central: https://humancellatlas.slack.com/archives/C9ENBPVNW/p1572525618104200

calvinnhieu: [Ticket](https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/matrix-service/419). Resolving.

matrix/common/etl/transformers/specimen_library.py
calvinnhieu: [Ticket](https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/matrix-service/420). Resolving.

matrix/common/constants.py
mckinsel: Oh yeah it will. I'll take out the extra one.

matrix/common/constants.py
mckinsel: Okay made this switch and loading the projects with the test data.

PR #424, opened 2019-11-08
Remove genes.tsv.gz from mtx format

calvinnhieu:


matrix/common/constants.py
```diff
@@ -504,10 +511,14 @@ class MetadataSchemaName(Enum):
...
 
-<h3><code>genes.tsv.gz</code></h3>
+<h3><code>features.tsv.gz</code></h3>
 <p>The gene metadata contains basic information about the genes in the count matrix.
 Each row is a gene, and each row corresponds to the same row in the expression mtx file.
 Note that <code>featurename</code> is not unique.</p>
+
+<h3><code>barcodes.tsv.gz</code></h3>
+<p>A list of cell barcodes corresponding to the columns found in matrix.mtx.gz.
+Note that barcodes may not be unique.</p>
```
calvinnhieu: @mckinsel What do you recommend we use instead to ensure uniqueness across projects? Will storing something other than barcodes in this file cause confusion?

matrix/common/constants.py
```diff
@@ -459,7 +459,7 @@ class MetadataSchemaName(Enum):
     MatrixFormat.MTX.value: """
```
calvinnhieu: Does this content need to be updated somewhere else too?

tests/functional/test_conversions.py
```diff
@@ -223,8 +222,19 @@ def test_mtx(self, mock_upload_method):
                 mtx_cells[row["cellkey"]] = row
 
             mtx_genes = collections.OrderedDict()
-            for row in csv.DictReader(io.StringIO(gzip.GzipFile(fileobj=io.BytesIO(
-                    mtx_output.read("test.mtx/genes.tsv.gz"))).read().decode()), delimiter='\t'):
+            for row in csv.DictReader(
+                    io.StringIO(gzip.GzipFile(fileobj=io.BytesIO(
+                    mtx_output.read("test.mtx/features.tsv.gz"))).read().decode()),
+                    delimiter='\t',
+                    fieldnames=["featurekey",
```
calvinnhieu: Required since `features.tsv.gz` per specification does not include headers. Will this be confusing to users given we're removing `genes.tsv.gz`?

calvinnhieu:
Closing until #425 is resolved.

PR #434, opened 2019-12-03
parse unique UUIDs from DB manifest in v0

calvinnhieu:
The DB has [updated their file manifest](https://github.com/DataBiosphere/azul/issues/1489) for which now may return duplicate UUIDs. This change handles these cases in order to prevent `resolved bundles in request do not match bundles available in matrix service` errors.

### DIFF
--- a/matrix/lambdas/daemons/v0/driver.py
+++ b/matrix/lambdas/daemons/v0/driver.py
@@ -1,3 +1,4 @@
+import collections
 import os
 import typing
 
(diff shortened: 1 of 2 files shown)


matrix/lambdas/daemons/v0/driver.py
```diff
@@ -161,7 +162,7 @@ def _parse_line(line: str) -> str:
             return tokens[0]
 
         lines = data.splitlines()[1:]
-        return list(map(_parse_line, lines))
+        return list(collections.OrderedDict.fromkeys(map(_parse_line, lines)))
```
Bento007: This avoids the needs for importing collection and is easier to understand.

```suggestion
        return list(set(map(_parse_line, lines)))
```

matrix/lambdas/daemons/v0/driver.py
```diff
@@ -1,3 +1,4 @@
+import collections
```
Bento007:
```suggestion
```

